PR #483, opened 2022-12-22
Don't log duplicate ids for image classification

franz101:

In the model logger check if id has been logged before.
If so skip logging it.

### DIFF
--- a/dataquality/loggers/model_logger/image_classification.py
+++ b/dataquality/loggers/model_logger/image_classification.py
@@ -76,10 +76,4 @@ class ImageClassificationModelLogger(TextClassificationModelLogger):
         with lock:
             self._filter_duplicate_ids()
 
-        if len(self.ids) != len(set(self.ids)):
-            raise ValueError(
-                f"Duplicate ids found in batch: {self.ids}. "
-                "Please make sure that each sample has a unique id."
-            )
-
         return super()._get_data_dict()
--- a/dataquality/loggers/model_logger/image_classification.py
+++ b/dataquality/loggers/model_logger/image_classification.py
@@ -10,6 +10,8 @@ from dataquality.loggers.model_logger.text_classification import (
 )
 from dataquality.schemas import __data_schema_version__
 from dataquality.schemas.split import Split
+from dataquality.utils.dq_logger import get_dq_logger
+from dataquality.utils.thread_pool import lock
 
 
 class ImageClassificationModelLogger(TextClassificationModelLogger):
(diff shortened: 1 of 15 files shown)


dataquality/loggers/model_logger/image_classification.py
```diff
@@ -33,3 +35,45 @@ def __init__(
...
+    def _filter_duplicate_ids(self) -> None:
+        """To avoid duplicate ids, when augmentation is used.
+        Filter out duplicate ids in the batch. This is done by keeping track of
+        the ids that have been observed in the current epoch in the config"""
+        observed_ids = self.logger_config.observed_ids[str(self.epoch)]
+        unique_ids = set(self.ids).difference(observed_ids)
+        observed_ids.update(unique_ids)
+        # If there are duplicate ids, filter out the duplicates
+        if len(self.ids) > len(unique_ids):
+            unique_indices = [list(self.ids).index(id) for id in unique_ids]
```
Ben-Epstein, line 48: calling `list()` and then `.index` inside this list comprehension is creating a `2n^2` complexity. Try creating the list of `self.ids` first, and create a dictionary if `id: index` so the lookup is `O(1)`

dataquality/loggers/model_logger/image_classification.py
```diff
@@ -33,3 +35,45 @@ def __init__(
...
+            if len(self.embs) > 0:
+                self.embs = np.array(self.embs)[unique_indices]
+            if len(self.probs) > 0:
+                self.probs = np.array(self.probs)[unique_indices]
+            if len(self.ids) > 0:
+                self.ids = np.array(self.ids)[unique_indices]
+
+    def _get_data_dict(self) -> Dict[str, Any]:
```
Ben-Epstein, line 56: you can refactor this a bit, to make it utilize the parent function a bit more
```suggestion
    def _get_data_dict(self) -> Dict[str, Any]:
        self._filter_duplicate_ids()
        if len(self.ids) != len(set(self.ids)):
            raise ValueError(
                f"Duplicate ids found in batch: {self.ids}. "
                "Please make sure that each sample has a unique id."
            )
        return super()._get_data_dict()
```

dataquality/loggers/model_logger/image_classification.py
franz101: great feedback. added this

dataquality/loggers/model_logger/image_classification.py
```diff
@@ -33,3 +35,51 @@ def __init__(
...
+            return super().write_model_output(model_output)
+
+    def _get_data_dict(self) -> Dict[str, Any]:
+        # Handle the binary case by converting it to 2-class classification
+        with lock:
+            self._filter_duplicate_ids()
+
+        if len(self.ids) != len(set(self.ids)):
+            raise ValueError(
+                f"Duplicate ids found in batch: {self.ids}. "
```
Ben-Epstein, line 81: do we need this anymore? We are filtering duplicate ids right above this

Ben-Epstein (approved):

anthonycorletti:
wasnt quite ready yet. have to double check all data structure operations in _filter_duplicate_ids :(

PR #487, opened 2023-01-12
UI changed LM default to 100%

Ben-Epstein:
we changed the default LM percentile to 100% (which is a dep percentile of 0), but never updated the dq client.


franz101 (commented):
👍

franz101 (approved):
👍

franz101 (approved):
Screenshot 2023-01-12 at 16 59 32


double checked and looks good to me

PR #489, opened 2023-01-13
Small/opti process df

Ben-Epstein:
small cleanups and speedups. The `get_labels` call for every task (in customer cases) were like 200 calls, because of 200 tasks. Pulling it like this is so much faster(100x)


dmallon (approved):

dataquality/metrics.py
```diff
@@ -658,8 +657,24 @@ def get_xray_cards(
...
+        if str(e) != "For multi-label runs, a task name must be provided":
+            raise e
+        # Multi-label case with no task provided. Get all tasks
+        pid, rid = api_client._get_project_run_id(project_name, run_name)
+        labels_object = f"{pid}/{rid}/labels/labels.json"
+        labels_path = object_store.download_file(
+            labels_object, "/tmp/labels.json", object_store.RESULTS_BUCKET_NAME
+        )
+        with open(labels_path) as f:
+            return json.loads(f.read())
```
elboy3, line 677: this would be a list of list of strings, right?

Ben-Epstein: yea it would be :( - the classic MLC conundrum. wdyt?

elboy3: it's fine but we should update the return type then

PR #491, opened 2023-01-18
Speed up base64 encoding step for CV data logging

rfriel:

When an image dataset is logged to Galileo, we convert the image data into a base64 encoded string.  Currently, what we do is

1. Load an image path into a PIL `Image`
2. Get the MIME type from the `Image`
3. Get bytes from the image by calling `Image.save`, using the same image format 
4. Combine the bytes and the MIME type and UTF-8 encode them to a string

This is wasteful, because we already have bytes on disk -- that is what we load in step 1.  If the image is lossy-compressed like JPEG, we may end up double-compressing it in step 3.  And we're run a JPEG decoder (step 1) and possibly encoder (step 3) when we don't need to.

Here I replace the above with

1. Load the image path to bytes
2. Infer the MIME type from the path using python stdlib's `mimetypes`. (If it can't infer a MIME type, fall back to using PIL `Image` as above)
4. Combine the bytes and the MIME type and UTF-8 encode them to a string

Speed comparison on 512x512 images from food-101 dataset:

- main branch: 89.15it/s
- this code: 5416.25it/s

I tested this manually in a notebook by running

```bash
pip install 'dataquality @ git+https://github.com/rungalileo/dataquality@feature/base64-speedup'
```

and then logging a small dataset.  The finished run is [here](https://console.dev.rungalileo.io/insights?dataframeColumns=confidence%3B%26data_error_potential%3B%26gold%3B%26level_0%3B%26path%3B%26pred%3B%26text&projectId=b4f8d2d1-a847-4541-8975-7cbe71eac7e8&runId=8256fecd-a307-4931-96a7-852cc8c99bb2&taskType=3)

Doesn't look like there are any automated tests for this code path.  EDIT: added one as `test_base64_image_logging`

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -1,6 +1,6 @@
 "dataquality"
 
-__version__ = "v0.8.13"
+__version__ = "v0.8.14"
 
 import os
 import warnings
--- a/dataquality/loggers/data_logger/image_classification.py
+++ b/dataquality/loggers/data_logger/image_classification.py
@@ -2,7 +2,6 @@ import os
 from typing import List, Optional, Union
 
 import pandas as pd
-from PIL import Image
 
 from dataquality.loggers.data_logger.base_data_logger import ITER_CHUNK_SIZE, MetasType
 from dataquality.loggers.data_logger.text_classification import (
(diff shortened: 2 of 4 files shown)


anthonycorletti (approved):

Ben-Epstein (changes requested):
nice! Can we add a quick test for these helpers?

rfriel:
> nice! Can we add a quick test for these helpers?

I added `test_base64_image_logging` -- is that the kind of thing you wanted?  It's a bit complicated... I wasn't sure how to write a simple test given that we have to deal with images and files here.

I wanted to turn the part that makes a synthetic image dataset into a shared utility, but it might be tricky because of the tempfile stuff, so I just added a TODO.

Ben-Epstein (approved):
this is great!

elboy3 (approved):

PR #492, opened 2023-01-20
Simplify the init function

elboy3:
Big init simplification! It was overly complicated

Screen Shot 2023-01-20 at 1 22 42 PM

### DIFF
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -2,7 +2,7 @@ import os
 import re
 import shutil
 import warnings
-from typing import Dict, Optional
+from typing import Dict, Optional, Tuple
 
 from pydantic.types import UUID4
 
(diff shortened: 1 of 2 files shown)


dataquality/core/init.py
```diff
@@ -21,43 +21,45 @@
...
-    ) -> Dict:
-        return api_client.get_project_run_by_name(project_name, run_name)
+        Returns:
+            Tuple[Dict, bool]: The project and a boolean indicating if the project
+            was created
+        """
+        project = api_client.get_project_by_name(project_name)
+        created = False
+        if not project:
+            created = True
```
asanyal, line 37: nit: should line 37 be after line 38?

dataquality/core/init.py
```diff
@@ -21,43 +21,45 @@
...
+    ) -> Tuple[Dict, bool]:
+        """Gets a run by name, or creates a new one if it doesn't exist.
+
+        Returns:
+            Tuple[Dict, bool]: The run and a boolean indicating if the run was created
+        """
+        run = api_client.get_project_run_by_name(project_name, run_name)
+        created = False
+        if not run:
+            created = True
```
asanyal, line 56: same as above. 56 should be after line 57.

dataquality/core/init.py
```diff
@@ -21,43 +21,45 @@
 BAD_CHARS_REGEX = r"[^\w -]+"
 
 
-class _Init:
-    def get_project_by_name_for_user(self, project_name: str) -> Dict:
-        return api_client.get_project_by_name(project_name)
+class InitManager:
```
Ben-Epstein, line 24: thank god, great rename

Ben-Epstein (commented):
This looks really great @elboy3, this code was long due for a cleanup. 

One concern here is that you removed the retry logic for the duplicate project creation. That issue still exists, I think we need to handle it accordingly. 

But, in the spirit of much cleaner code, I have a much simpler idea than what is currently written/being removed.

In the `get_or_create_project` function, let's simply tack on a `@retry` with a retry limit of 1, a backoff of 500ms, and an exception matcher for `GalileoException`

So (with the tenacity library) that would just look like
```
from tenacity import (
    retry,
    retry_if_exception_type,
    stop_after_attempt,
    wait_fixed,
)

@retry(stop=stop_after_attempt(1), wait=wait_fixed(0.5), retry_if_exception_type(GalileoException))
def get_or_create_project
```

dataquality/core/init.py
```diff
@@ -21,43 +21,45 @@
...
+    def get_or_create_project(
```
Ben-Epstein, line 25: let's keep your clean layout, but still capture that duplicate project name error by retrying this function once on a GalileoException

```suggestion
    @retry(stop=stop_after_attempt(1), wait=wait_fixed(0.5), retry_if_exception_type(GalileoException))
    def get_or_create_project(
```

elboy3:
> This looks really great @elboy3, this code was long due for a cleanup.
> 
> One concern here is that you removed the retry logic for the duplicate project creation. That issue still exists, I think we need to handle it accordingly.
> 
> But, in the spirit of much cleaner code, I have a much simpler idea than what is currently written/being removed.
> 
> In the `get_or_create_project` function, let's simply tack on a `@retry` with a retry limit of 1, a backoff of 500ms, and an exception matcher for `GalileoException`
> 
> So (with the tenacity library) that would just look like
> 
> ```
> from tenacity import (
>     retry,
>     retry_if_exception_type,
>     stop_after_attempt,
>     wait_fixed,
> )
> 
> @retry(stop=stop_after_attempt(1), wait=wait_fixed(0.5), retry=retry_if_exception_type(GalileoException))
> def get_or_create_project
> ```

@Ben-Epstein Good feedback. I intentionally removed it, but open to discussion. With the `get_or_create` duplicate request race condition seems extremely unlikely. How would it even happen, two people try to init with the same project name at the exact same time? Is this something we have run into before? 

Maybe even if it's extremely unlikley it's still worth having the retry. But wanted to get some more context on if this has ever happened before

Ben-Epstein:
@elboy3 in "normal" conditions it can be unlikely, but in the case of automated job starting, such as dojo, it's extremely common.

Dojo starts many jobs at once, so if you start a `dojo run` process with a new project that doesn't exist, a bunch will fail for this exact reason. That was the original reason we added the check

Ben-Epstein (approved):
great!

PR #493, opened 2023-01-23
Feature/structured classification

elboy3:

Closes:
https://app.shortcut.com/galileo/story/3454/dq-init-structured-data-run
https://app.shortcut.com/galileo/story/3516/dq-log-structured-dataset
https://app.shortcut.com/galileo/story/3455/dq-log-structured-samples

[sc-3454] [sc-3516] [sc-3455]

Pytest and jupyter notebooks!

***Demonstration***

Adding functions `log_structured_samples` and `log_structured_dataset` as entrypoints for users to access Galileo's new Structured Data modality.


dataquality/core/log.py
```diff
@@ -142,6 +146,95 @@ def log_image_dataset(
...
+            "my_labels": ["A", "A", "B"],
+            "my_id": [1, 2, 3],
+            "sample_quality": [5.3, 9.1, 2.7]
+        }
+
+    In the invalid case, use `dq.log_data_samples`:
+        meta = {"sample_quality": d["sample_quality"]}
+        dq.log_data_samples(
+            texts=d["my_text"], labels=d["my_labels"], ids=d["my_ids"], meta=meta
+        )
```
Ben-Epstein, line 211: this seems all about NLP data. I'd keep "Invalid example" but update this to a tabular example, and change it to `dq.log_structured_samples`

dataquality/core/log.py
```diff
@@ -142,6 +146,95 @@ def log_image_dataset(
...
+        probs=probs,
+        split=split,
+        inference_name=inference_name,
+    )
+
+
+@check_noop
+def log_structured_dataset(
+    dataset: pd.DataFrame,
+    probs: np.ndarray,
```
Ben-Epstein, line 180: some people _may_ have the `probs` as a column in the pandas df.. Not necessarily, but it might be worth considering supporting that now or in the future, instead of requiring a separate numpy array.

You can have a column with lists in a pandas df, its just a little odd

dataquality/core/log.py
```diff
@@ -142,6 +146,95 @@ def log_image_dataset(
...
+    assert all(
+        [config.task_type, config.current_project_id, config.current_run_id]
+    ), "You must call dataquality.init before logging data"
+    data_logger = get_data_logger()
+    assert isinstance(data_logger, StructuredClassificationDataLogger), (
+        "This method is only supported for structured data tasks. "
+        "You must call dq.init('structured_classification') to use this method."
+    )
+    data_logger.log_structured_dataset(
+        dataset=dataset,
```
Ben-Epstein, line 230: do you want to validate that the dataset is a pandas df?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,274 @@
...
+class StructuredClassificationDataLogger(BaseGalileoDataLogger):
+    __logger_name__ = "structured_classification"
+    logger_config = structured_classification_logger_config
+
+    def validate(self) -> None:
+        """Validates the input data before logging
+
+        Validates:
+            - Split is set (TODO)
+            - Inference name is set if split is inference (TODO)
```
Ben-Epstein, line 33: shouldnt the base logger validate this already?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,274 @@
...
+            - The length of the data, labels, and probs are the same
+            - The number of feature names is the same as the number of features
+        """
+        assert len(self.X) == len(self.y) == len(self.probs)
+        assert len(self.feature_names) == len(self.X[0])
```
Ben-Epstein, line 38: I'd add an error message for each of these that is user friendly

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def log_structured_dataset(
+        self,
+        dataset: pd.DataFrame,
+        probs: np.ndarray,
+        label: str,
+        split: str = None,
+        inference_name: str = None,
+    ) -> None:
+        self.y = dataset[label].values
```
Ben-Epstein, line 65: idk if this is true for pandas, but `.values` may be risky. Pandas is built on pyarrow, so it's possible you might get back a pyarrow ListArray, or a numpy array. If that's the case, I'd be explicit and use `tolist` or `to_numpy` (or whatever the api is for pandas)

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,274 @@
...
+        data = {
+            "id": ids,
+            "gold": self.y,
+            "prob": self.probs,
+            "split": [self.split] * num_samples,
+            "data_schema_version": [__data_schema_version__] * num_samples,
+            **features,
+        }
+        if self.split == Split.inference:
+            data.update(inference_name=self.inference_name)
```
Ben-Epstein, line 142: does this not need to be `data.update(inference_name=[self.inference_name]*num_samples)` like split?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,274 @@
...
+        dataframes = self.separate_dataframe(df)
+
+        prob = dataframes.prob
+        data_df = dataframes.data
+
+        # Upload data and prob dfs to minio
+        for data_folder, df_obj in tqdm(
+            zip(DATA_FOLDERS, [data_df, prob]),
+            total=len(DATA_FOLDERS),
+            desc="Uploading Data",
```
Ben-Epstein, line 192: You have an outer tqdm bar in `upload` with the same message.

I would
1. make the description `Uploading {split} data` since you know the split at this point
2. Make this bar disappear after upload (we have that in other places, I think the api is like `leave=False` or something

OR you might just not need this inner tqdm bar at all. You are uploading relatively small amounts of data (i presume), and we have the upload progress bar in `create_project_run_object_from_df` - so there may be enough live user feedback to not warrent this inner progress bar

tests/conftest.py
```diff
@@ -64,7 +64,7 @@ def stunted_get(url: str) -> MockResponse:
 @pytest.fixture(scope="function")
 def cleanup_after_use() -> Generator:
     for task_type in list(TaskType):
-        dataquality.get_model_logger(task_type).logger_config.reset()
+        dataquality.get_data_logger(task_type).logger_config.reset()
```
Ben-Epstein, line 67: i see why you did this, but for continuity sake I feel like you should create a `StructuredDataModelLogger` that inherits from `BaseModelLogger` and just doens't do anything...

Ben-Epstein (commented):
This all looks great. 

You implemented this following the paradigms of the other tasks which is nice for consistency and (i think) should make this scalable to really big datasets in theory.

The one thing though, is that because in structured we fundamentally log all the data at once, I'm not sure the complexity is necessarily worth it? I'll definitely leave that up to you, but consider the flow of data currently:

1. Get pandas dataframe from user
2. Parse it into our `DataLogger` class 
3. Validate the data
4. Parse that data into a dictionary (which was essentially the same as the pandas df `.to_records()`)
5. Write that to disk as an hdf5 file
6. Read it from disk into a vaex dataframe 
7. Upload that vaex df into 2 parts (probs, data)

It's a bit redundant in the passing around of the data. For example, you are validating that all of the features have the same length, but since it started in a pandas dataframe you don't actually need to do that.

A potentially simpler flow would be to simply
1. Get the pandas df from the user
2. Validate the pandas df (not even sure really what needs to be validated)
3. Create a vaex df in-memory `df = vaex.from_pandas_df(pdf)`
4. Upload that vaex df into 2 parts (probs, data)

You certainly don't need to do that, but (unless I'm missing something), that would accomplish the same thing

dataquality/loggers/model_logger/base_model_logger.py
```diff
@@ -128,16 +128,6 @@ def write_model_output(self, data: Dict) -> None:
...
-        # Inference split must have inference name
-        if self.split == Split.inference and self.inference_name is None:
-            if self.logger_config.cur_inference_name is not None:
-                self.inference_name = self.logger_config.cur_inference_name
-            else:
-                raise GalileoException(
-                    "For inference split you must either log an inference name "
-                    "or set it before logging. Use `dataquality.set_split` to set"
-                    "inference_name"
-                )
```
elboy3, line 131: moved to base logger

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+        error_msg = (
+            "must be set before setting probs. Try calling `log_structured_dataset`"
+        )
+        assert self.model is not None, f"Model {error_msg}"
+        assert self.dataset is not None, f"Dataset {error_msg}"
+
+        input_data = self.dataset.copy()
+        # Drop gold and id columns if they exist
+        input_data = input_data.drop(["gold", "id"], axis=1, errors="ignore").to_numpy()
+        self.probs = self.model.predict_proba(input_data)
```
asanyal, line 100: Quick note to measure the perf of this line of code. In general would be great to do some basic tests and measure the latency of `predict_proba` + how it varies with dataset sizes.

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+        objectstore.create_project_run_object_from_df(
+            data_df, f"{object_base_path}/data/data.hdf5"
+        )
+        objectstore.create_project_run_object_from_df(
+            probs_df, f"{object_base_path}/prob/prob.hdf5"
+        )
+
+    def _get_dfs(self) -> Tuple[DataFrame, DataFrame]:
+        """Returns data and probs as vaex DataFrames"""
+        df = vaex.from_pandas(self.dataset)
```
asanyal, line 167: Just for my knowledge -- do we have a sense of how expensive these dataframe conversions calls are? My guess would be it's cheap, given it's not copying any data over, but want to confirm this.

dataquality/core/log.py
```diff
@@ -1,6 +1,8 @@
 from typing import Any, Dict, List, Optional, Type, Union
 
 import numpy as np
+import pandas as pd
+import xgboost as xgb
```
franz101, line 5: Lazy import?

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
...
+    feature_names: List[str],
+    *,
+    y: Optional[np.ndarray] = None,
+    split: Optional[Split] = None,
+    inference_name: Optional[str] = None,
+) -> None:
+    """Log a split with raw numpy arrays for structured data
+
+    Example:
+    .. code-block:: python
```
franz101, line 163: love the code block :)

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
     )
 
 
+@check_noop
+def log_structured_samples(
+    model: xgb.XGBClassifier,
```
franz101, line 152: in some cases a str can do the job.
def func2(self, some_param: 'Main'):

franz101: A couple of questions:
- Why would we pass in a xgboost classifier?

dataquality/loggers/base_logger.py
```diff
@@ -74,17 +74,23 @@ def __init__(self) -> None:
...
         """Returns the path to the output directory for the current run
 
         Example:
             /Users/username/.galileo/logs/proj-id/run-id
         """
-        return (
-            f"{BaseGalileoLogger.LOG_FILE_DIR}/{config.current_project_id}/"
-            f"{config.current_run_id}"
-        )
+        return f"{BaseGalileoLogger.LOG_FILE_DIR}/{self.proj_run}"
```
franz101, line 93: 👍

dataquality/loggers/base_logger.py
```diff
@@ -102,6 +108,22 @@ def split_name(self) -> str:
             split = f"{split}_{self.inference_name}"
         return str(split)
```
franz101, line 109: str(str())?

dataquality/loggers/base_logger.py
```diff
@@ -102,6 +108,22 @@ def split_name(self) -> str:
...
+        If the split is inference, it will return the name of the inference
+        run after the split name
+
+        Example:
+            training
+            inference/inf-name1
+        """
+        split = self.split
+        if split == Split.inference:
+            split = f"{split}/{self.inference_name}"
```
franz101, line 124: for complex paths I can recommend the Pathlib

dataquality/loggers/base_logger.py
```diff
@@ -123,6 +145,18 @@ def set_split_epoch(self) -> None:
...
                 )
+
+        # Inference split must have inference name
+        if self.split == Split.inference and self.inference_name is None:
+            if self.logger_config.cur_inference_name is not None:
+                self.inference_name = self.logger_config.cur_inference_name
+            else:
+                raise GalileoException(
+                    "For inference split you must either log an inference name "
+                    "or set it before logging. Use `dataquality.set_split` to set "
```
franz101, line 156: Some syntactic sugar:
```python
if self.split == Split.inference and not self.inference_name:
    self.inference_name = self.logger_config.cur_inference_name or raise GalileoException(
        "For inference split, either log an inference name "
        "or set it with `dataquality.set_split`. Current inference name is not set."
    )
```

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+from dataquality.clients.objectstore import ObjectStore
+from dataquality.loggers.data_logger.base_data_logger import BaseGalileoDataLogger
+from dataquality.loggers.logger_config.structured_classification import (
+    structured_classification_logger_config,
+)
+from dataquality.schemas import __data_schema_version__
+from dataquality.schemas.split import Split
+
+
+class StructuredClassificationDataLogger(BaseGalileoDataLogger):
```
franz101, line 18: General question, does StructuredClassificationDataLogger always need a model? If not would XGB as subset?

dataquality/loggers/data_logger/structured_classification.py
elboy3: @rfriel this could be a good place to run a few further experiments and confirm

dataquality/loggers/data_logger/structured_classification.py
elboy3: Extremely quick with vaex. Subsecond even on large datasets

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(
+        self,
+        model: Optional[xgb.XGBClassifier] = None,
+        X: Optional[np.ndarray] = None,
+        y: Optional[np.ndarray] = None,
+        probs: Optional[np.ndarray] = None,
+    ) -> None:
+        super().__init__()
+        self.model: xgb.XGBClassifier = (
+            model if model is not None else xgb.XGBClassifier()
```
franz101, line 31: or is your friend:
model or xgb.XGBClassifier()

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+        self.X: np.ndarray = X if X is not None else np.array([])
+        self.y: np.ndarray = y if y is not None else np.array([])
+        self.probs: np.ndarray = probs if probs is not None else np.array([])
+
```
franz101, line 36: sadly the same simplification doesn't work here:
`The truth value of an empty array is ambiguous`

franz101: Make parameter X np.array([]) by default?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+            - The length of the data and probs are the same
+        """
+        super().validate()
+        assert len(self.dataset) == len(self.probs), (
+            "Data and probs are not the same length. "
+            f"Data: {len(self.dataset)}, Probs: {len(self.probs)}"
+        )
+
+    @classmethod
+    def create_dataset_from_samples(
```
franz101, line 52: create_df_from_samples

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+        self.model = model
+        self.split = split
+        self.inference_name = inference_name
+        self.dataset = self.create_dataset_from_samples(X, y, feature_names)
+        self.log()
+
+    def log_structured_dataset(
+        self,
+        model: xgb.XGBClassifier,
+        dataset: pd.DataFrame,
```
franz101, line 120: curios about dataframe = dataset.
no lists/numpy arrays here?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+
+        if y is not None and y.any():
+            assert len(X) == len(y), "Data and labels are not the same length. "
+            f"Data: {len(X)}, Labels: {len(y)}"
+            dataset["gold"] = y
+
+        return dataset
+
+    def set_probs(self) -> None:
+        """Sets the probs attribute for the class
```
franz101, line 85: what is the probs attribute?

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,197 @@
...
+        self.dataset = dataset
+        self.dataset.rename(columns={label: "gold"}, inplace=True)
+        self.split = split
+        self.inference_name = inference_name
+        self.log()
+
+    def log(self) -> None:
+        """Writes input data to disk in .galileo/logs
+
+        Note that unlike unstructured data, we don't batch data here, we write
```
franz101, line 135: why is the difference? I actually like this for structured data as well. otherwise init needs to be called all the time.

dataquality/loggers/model_logger/structured_classification.py
```diff
@@ -0,0 +1,15 @@
+from typing import Dict
+
+from dataquality.loggers.logger_config.structured_classification import (
+    structured_classification_logger_config,
+)
+from dataquality.loggers.model_logger.base_model_logger import BaseGalileoModelLogger
+
+
+class StructuredClassificationModelLogger(BaseGalileoModelLogger):
+    # NOTE: Structured Classification doesn't require a model logger
```
franz101, line 10: because the model is not logging each time?

franz101 (commented):
a couple of questions before I approve:
```why not dq.log_dataset(data)
and dq.log_model(model)```
and are we sure xgboost is not another requirement when we import from the core module?

🙏

File: dataquality/core/log.py (hunk 13 above)
Comment (elboy3) [reply dataquality/core/log.py:5]: why?

File: dataquality/core/log.py (hunk 15 above)
Comment (elboy3) [reply dataquality/core/log.py:152]: The user needs to log their model for structured data. This will allow us to calculate things like dep, feature importance, etc.

File: dataquality/core/log.py (hunk 13 above)
Comment (elboy3) [reply dataquality/core/log.py:5]: What do you think about the follpwing instead of lazy import 

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from xgboost import XGBClassifier
```

File: dataquality/loggers/base_logger.py (hunk 17 above)
Comment (elboy3) [reply dataquality/loggers/base_logger.py:109]: 🤷 this is just how it was already. i can remove

File: dataquality/loggers/base_logger.py (hunk 19 above)
Comment (elboy3) [reply dataquality/loggers/base_logger.py:156]: 😋

File: dataquality/loggers/data_logger/structured_classification.py (hunk 22 above)
Comment (elboy3) [reply dataquality/loggers/data_logger/structured_classification.py:36]: i think setting default to none is better, since the default param in python i mutable. but will do your suggestion for model above!

File: dataquality/loggers/data_logger/structured_classification.py (hunk 25 above)
Comment (elboy3) [reply dataquality/loggers/data_logger/structured_classification.py:85]: what do you mean?

File: dataquality/loggers/data_logger/structured_classification.py (hunk 26 above)
Comment (elboy3) [reply dataquality/loggers/data_logger/structured_classification.py:135]: init only needs to be called once if I'm thinking about it right! 

basically with unstructured data we would have to batch upload because 1) epochs and 2) their texts may not fit into memory.

We don't have epochs or memory concerns (for now) in structured data

File: dataquality/loggers/model_logger/structured_classification.py (hunk 27 above)
Comment (elboy3) [reply dataquality/loggers/model_logger/structured_classification.py:10]: model outputs are not logged at all :) 

we get the model prob outputs from their model which is already logged

**Comment by elboy3** [comment]
@franz101 great review! I think it's better to just do one log `log_structured` than have them do 2 back to back logs. WDYT?

File: dataquality/loggers/base_logger.py (hunk 19 above)
Comment (elboy3) [reply dataquality/loggers/base_logger.py:156]: wait. this doesn't work :( 

```
>>> a = 5
>>> b = ""
>>> c = a or raise Exception("yoo")
  File "<stdin>", line 1
    c = a or raise Exception("yoo")
             ^
SyntaxError: invalid syntax
```

File: dataquality/core/log.py (hunk 13 above)
Comment (franz101) [reply dataquality/core/log.py:5]: That's cool, i didn't know it existed

File: dataquality/loggers/data_logger/structured_classification.py (hunk 11 above)
Comment (rfriel) [reply dataquality/loggers/data_logger/structured_classification.py:98]: I've done some experiments.  Briefly, yes, we can assume predict_proba is very fast (under 1 second for reasonable inputs).

In particular, for reasonably sized datasets and models, running `predict_proba` on the training data takes less than 1% as much time as it took to `fit` the model on the same data.

So if the user has the patience to fit the model, we're not adding much overhead on top of that.

File: dataquality/loggers/data_logger/structured_classification.py [hunk 28]
```diff
@@ -0,0 +1,195 @@
...
+        self,
+        model: xgb.XGBClassifier,
+        dataset: pd.DataFrame,
+        label: Optional[str],
+        split: str = None,
+        inference_name: str = None,
+    ) -> None:
+        self.model = model
+        self.dataset = dataset
+        self.dataset.rename(columns={label: "gold"}, inplace=True)
```
rfriel, line 125: Will this modify the dataframe the user passes as `dataset`?

That seems potentially confusing.

Also, if something fails in `log()`, and we call this function again to re-try, it will now fail because there's no column with the original label name.

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,195 @@
...
+        self.split = split
+        self.inference_name = inference_name
+        self.dataset = self.create_dataset_from_samples(X, y, feature_names)
+        self.log()
+
+    def log_structured_dataset(
+        self,
+        model: xgb.XGBClassifier,
+        dataset: pd.DataFrame,
+        label: Optional[str],
```
rfriel, line 119: We need label to not be None.

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+        assert self.model is not None, f"Model {error_msg}"
+        assert self.dataset is not None, f"Dataset {error_msg}"
+
+        input_data = self.dataset.copy()
+        # Drop gold and id columns if they exist
+        input_data = input_data.drop(["gold", "id"], axis=1, errors="ignore").to_numpy()
+        self.probs = self.model.predict_proba(input_data)
+
+    def log_samples(
```
rfriel, line 100: It would be nice if this called `log_structured_dataset`, to avoid having two code paths that can diverge.  Like

```python
        self.model = model
        self.split = split
        self.inference_name = inference_name
        self.dataset = self.create_dataset_from_samples(X, y, feature_names)
        self.log_structured_dataset(
                model=self.model, 
                dataset=self.dataset, 
                label="gold",
                split=self.split,
                inference_name=self.inference_name,
        )
```

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -0,0 +1,195 @@
...
+        Assumes dataset is input data X, with optional "id" and "gold" columns
+            that are removed
+        """
+        error_msg = (
+            "must be set before setting probs. Try calling `log_structured_dataset`"
+        )
+        assert self.model is not None, f"Model {error_msg}"
+        assert self.dataset is not None, f"Dataset {error_msg}"
+
+        input_data = self.dataset.copy()
```
rfriel, line 95: I don't think the copy is necessary.  `input_data.drop` on the next line isn't inplace, it returns a modified copy.

tests/loggers/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+            "X": X_test,
+            "dataset": test_ds,
+        },  # It's fine for inference data to be test data
+        "inf2": {"X": X_test, "dataset": test_ds},
+    }
+
+
+@pytest.fixture(scope="module")
+def fit_xgboost(sc_data: Dict) -> xgb.XGBClassifier:
+    model = xgb.XGBClassifier(objective="multi:softprob", random_state=42)
```
rfriel, line 44: FYI, `XGBClassifier` will automatically set `objective="multi:softprob"` if passed multiclass data.

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
...
+    data_logger = get_data_logger()
+    assert isinstance(data_logger, StructuredClassificationDataLogger), (
+        "This method is only supported for structured data tasks. "
+        "You must call dq.init('structured_classification') to use this method."
+    )
+    assert isinstance(
+        model, xgb.XGBClassifier
+    ), "Logging structured data currently only supports XGBoost for classification."
+
+    data_logger.log_structured_dataset(
```
rfriel, line 272: We should probably also check whether the model has been fitted.  Something like

```
from sklearn.utils.validation import check_is_fitted
    [...]
    assert check_is_fitted(
        model
    ), "This model has not been fitted.  Fit the model before calling log_structured_dataset."
```

dataquality/__init__.py
```diff
@@ -35,6 +35,8 @@
     log_dataset,
     log_image_dataset,
     log_model_outputs,
+    log_structured_dataset,
```
rfriel, line 38: These names are potentially confusing -- they sound like they're just about the data, but they're really about the data together with a fitted model.

We're treat it as "data logging" internally, but that's an implementation detail.

IMO the user-facing name should express "this is the function that logs everything to Galileo, if you're doing structured classification."

The "samples" vs. "dataset" nomenclature is also confusing.  Makes it sounds like if you're logging "samples," you're only logging a subset of the data... or something.

So maybe 

- `log_structured_samples` --> `log_structured_classification_from_arrays`
- `log_structured_dataset` --> `log_structured_classification_from_df`

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
...
+        feature_names=feature_names,
+        split=split,
+        inference_name=inference_name,
+    )
+
+
+@check_noop
+def log_structured_dataset(
+    model: xgb.XGBClassifier,
+    dataset: pd.DataFrame,
```
rfriel, line 220: IMO, we should accept separate `X` and `y` arrays in this function, where `X` is a dataframe.

`XGBClassifier.fit(X, y)` can accept a pandas dataframe as `X`.  If you have a dataframe already, this is the most convenient way to fit the model.

However, you can't do `XGBClassifier.fit(df)` where df contains both X and y.

At the point where the user calls our logger, they have just done 

```python
# X: pd.DataFrame or np.ndarray
# y: np.ndarray
model.fit(X, y)
```

We can assume they have these two objects, X and y.  They might not have a combined dataframe with both X and y, and if so, it's awkward to ask them to create it.

This would also mean we don't need to ask the user for the (arbitrary) name that they assigned to the label in the combined df (the `label` arg).

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
...
+
+        data = load_wine()
+        X = data.data
+        y = data.target
+        feature_names = data.feature_names
+
+        model = xgb.XGBClassifier()
+        model.fit(X, y)
+
+        dq.log_dataset(model, X, feature_names, y=y, split="training")
```
rfriel, line 176: Shouldn't this be `dq.log_structured_samples`?  And below on line 179

dataquality/core/log.py
```diff
@@ -142,6 +147,137 @@ def log_image_dataset(
...
+        X = data.data
+        y = data.target
+        feature_names = data.feature_names
+
+        model = xgb.XGBClassifier()
+        model.fit(X, y)
+
+        dataset = pd.DataFrame(X, columns=feature_names)
+        dataset["label"] = y
+        dq.log_dataset(model, dataset, label="label", split="training")
```
rfriel, line 244: Shouldn't this be `dq.log_structured_dataset`?  And below on line 248

rfriel (commented):
Left some comments.  Mostly looks great!

Most comments were nits.  The important things are

- We could have more intuitive names for the user-facing functions
- The dataframe-logging function should have separate X and y arguments, like the numpy-logging one does

dataquality/loggers/data_logger/structured_classification.py
elboy3: As is now it will, good call. Will need to make sure to copy any pandas DF that we want to modify

dataquality/__init__.py
```diff
@@ -126,11 +125,10 @@ def set_console_url(console_url: str = None) -> None:
     "set_split",
     "set_epoch_and_split",
     "set_console_url",
+    "log",
```
rfriel, line 128: The user-facing name should be something that won't seem relevant if you're doing deep learning.

Possibilities:

- `dq.log_structured`
- `dq.log_xgboost`
  - if/when we make it explicit that we support general sklearn models, could be replaced with `dq.log_sklearn`
- or maybe create a new module under `dq`, like `dq.structured.log` or`dq.xgboost.log`

I much prefer `log_xgboost` to `log_structured`, since it's not obvious from context what "structured" means.

The last one seems like a good direction long-term when we have lots of modalities, to prevent `dq` from having a million log_whatever functions.  But maybe not in this PR, up to you.

dataquality/core/log.py
```diff
@@ -173,85 +177,40 @@ def log_structured_samples(
...
-    Example:
+    Example with pandas DataFrame:
     .. code-block:: python
 
         import xgboost as xgb
         from sklearn.datasets import load_wine
 
         data = load_wine()
-        X = data.data
+        df = pd.DataFrame(data.data, columns=data.feature_names)
```
rfriel, line 194: I would suggest writing this example like this:

```python
         X, y = load_wine(as_frame=True, return_X_y=True)

        model = xgb.XGBClassifier()
        model.fit(X, y)

        dq.log(model, X=df, y=y, split="training")
```

This is how I've been using the sklearn loaders in my experiments.

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -21,130 +23,99 @@ class StructuredClassificationDataLogger(BaseGalileoDataLogger):
...
+        if isinstance(self.X, np.ndarray):
+            assert self.feature_names is not None, (
+                "feature_names must be provided when logging X as a numpy array. "
+                "If X is a pandas DataFrame, feature_names will be inferred from the "
+                "column names."
+            )
+            assert len(self.X[0]) == len(
+                self.feature_names
+            ), "X and feature_names must have the same number of features"
+            self.X = (
```
rfriel, line 90: This is inside `if isinstance(self.X, np.ndarray):`, so we can remove the if/else part.

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -162,12 +133,15 @@ def log(self) -> None:
...
+        assert isinstance(self.X, pd.DataFrame), (
+            "X must be a pandas DataFrame. " f"X is currently a {type(self.X)}"
+        )
+        df = vaex.from_pandas(self.X)
         n_rows = len(df)
         ids = (
             df.id.to_numpy()
             if "id" in df.get_column_names()
-            else self.dataset.index.to_numpy()
+            else self.X.index.to_numpy()
```
rfriel, line 144: Pandas dataframes can have duplicates in their indices, and this often arises in practice.  But we can't handle duplicate IDs.

So we should either 

- check if the index has duplicates (eg `self X.index.duplicated().any()`), and if it does, fall back to `np.arange(len(self.X))` or something
- Or, always use `np.arange(len(self.X))` instead of the index if there's no ID column.

It's possible that the user has put meaningful, unique IDs in their dataframe index, and if so I guess we might want to preserve those...

But this behavior isn't documented, which means it's really bad if we fail because of it.  The user doesn't know we're doing this, and we're not telling them that their df has to have a unique index.

rfriel (changes requested):
Much cleaner/simpler this way!

Left some comments.

dataquality/loggers/data_logger/structured_classification.py
elboy3: id is used on our end to match columns in the data df and the probs df in the VaexService. So the ids don't really matter and the user will never see them. I think we should just go with `np.arange(len(self.X))`

dataquality/__init__.py
elboy3: I'm going to keep it simple with `log_xgboost` for now, but i definitely agree the future direction. 

as we add new models, requirements, modalities to structured i think the naming conventions will be more clear to us, and it will be easy to change

rfriel (approved):
💯

PR #494, opened 2023-01-25
Support HF CV datasets + non-path image columns

rfriel:

Adds support for logging Huggingface CV datasets with `dq.log_image_dataset`, without first converting them to pandas.

Relatedly, adds an argument `imgs_colname` to `dq.log_image_dataset`.

- `imgs_location_colname`: column containing file paths to images
- `imgs_colname`: column containing the images themselves

Supporting HF is not especially important for "real" (eg commercial) use cases, where we expect to get image paths, possibly in a cloud store.  However, it's useful for demos and example notebooks, and for internal testing.  These cases ought to "just work" without needing a bunch of manual steps to work around HF.

## Details

CV datasets in HF are inconvenient to work with in our current setup for data logging.

Right now, we only support passing the file paths of images in data logging.  However, HF datasets typically works by downloading an archive file (`.tar` or `.tar.gz`) and [reading it directly as an archive](https://huggingface.co/docs/datasets/v2.8.0/package_reference/builder_classes#datasets.DownloadManager.iter_archive), without ever unpacking it into separate files on the user's file system.

HF datasets will read these archives and give us the _data_ contained in them, but if we want file paths, we have to manually (and redundantly) save the data back to disk somewhere and then read it in again.  Avoiding this waste of time and disk space was a motivation for introducing `imgs_colname`.

### Images in HF

In a loaded HF dataset, images are expressed as [`datasets.Image`](https://huggingface.co/docs/datasets/v2.8.0/package_reference/main_classes#datasets.Image) features.  When accessed by the user, these will look one of two ways, depending on the `decode` setting of the `datasets.Image`feature:

- `decode=True`: a PIL image, loaded from disk
- `decode=False`: a dict with keys
  - `bytes`: the raw data from the file on disk
  - `path`: some sort of path.

In my experience, the `path` field is used inconsistently across HF CV datasets.)  Often it's `None`.  If not `None`, it's generally a relative path within the archive, like `images/train/0261.jpg`.  

I've also seen one case (the beans dataset) with the opposite pattern: `path` is supplied (and is an absolute path we can load), but `bytes` is None.

### Logging HF images

For dataset logging, we need two pieces of info for each image:

- the raw data, which we'll base64 encode
- a MIME type expressing the image format, eg JPEG

The code here tries to determine both of these as efficiently as possible, starting with the `path` and `bytes` information given for an example.

- _If both `path` and `bytes` are given_:
  - Get the raw data from `bytes`.
  - Use the file extension in `path` to guess the MIME type.
- _If `bytes` is given but `path` is not (or we can't guess the MIME type from `path`)_:
  - Get the raw data from `bytes`.
  - For the MIME type, load the raw data into an image with PIL, and call `get_format_mimetype()` on the PIL image.
- _If `path` is given but `bytes` is not (beans dataset):_
  - Get the raw data by reading the file at `path` (without decoding it as an image).
  - Use the file extension in `path` to guess the MIME type.

The first two cases are handled by `_bytes_to_b64_str`, the last one by `_img_path_to_b64_str`.

I've tried this on a variety of datasets, including MNIST, Imagenet, Tiny-Imagenet, food101, CIFAR-10, CIFAR-100, and beans.

### `imgs_colname` in the pandas case

Now that we have `imgs_colname`, I've added support for it in the case where you pass a pandas dataframe.

In this case, I expect a dataframe column where the entries are PIL images.

### Note on `imgs_dir`

I changed `imgs_dir` from a required to an optional argument of `dq.log_image_dataset`.

Why: it's only meaningful when we have file paths, they are relative paths, and the user knows what they are relative to.

In the HF dataset case, we typically don't have paths, or otherwise have absolute paths (beans).

In the pandas case, the user may give us absolute paths.  Asking for `imgs_dir` when you have absolute paths is confusing and unnecessary.

Added `test_hf_image_dataset`, which tests that we can pass an HF CV dataset directly to `dq.log_image_dataset` without errors.

This only tries one dataset, and hence only tests one particular code path among the various ones added here.

I'm not sure how extensively we should test this stuff, given that HF datasets are not our main (commercial) focus for CV.

Maybe it'd be nice to ensure we can handle a few very standard datasets -- maybe MNIST and CIFAR-10?  Could add those.

### DIFF
--- a/dataquality/loggers/data_logger/image_classification.py
+++ b/dataquality/loggers/data_logger/image_classification.py
@@ -2,7 +2,6 @@ import os
 from typing import List, Optional, Union
 
 import pandas as pd
-from PIL import Image
 
 from dataquality.loggers.data_logger.base_data_logger import ITER_CHUNK_SIZE, MetasType, DataSet
 from dataquality.loggers.data_logger.text_classification import (
@@ -13,7 +12,7 @@ from dataquality.loggers.logger_config.image_classification import (
     image_classification_logger_config,
 )
 from dataquality.schemas.split import Split
-from dataquality.utils.cv import _img_to_b64_str
+from dataquality.utils.cv import _img_path_to_b64_str
 
 
 class ImageClassificationDataLogger(TextClassificationDataLogger):
@@ -51,7 +50,7 @@ class ImageClassificationDataLogger(TextClassificationDataLogger):
         meta: Optional[List[Union[str, int]]] = None,
     ) -> None:
         dataset["text"] = dataset[imgs_location_colname].apply(
-            lambda x: _img_to_b64_str(img=Image.open(os.path.join(imgs_dir, x)))
+            lambda x: _img_path_to_b64_str(img_path=os.path.join(imgs_dir, x))
         )
         self.log_dataset(
             dataset=dataset,
--- a/dataquality/loggers/data_logger/image_classification.py
+++ b/dataquality/loggers/data_logger/image_classification.py
@@ -6,7 +6,6 @@ from PIL.Image import Image
 
 from dataquality.exceptions import GalileoException
 from dataquality.loggers.data_logger.base_data_logger import (
-    ITER_CHUNK_SIZE,
     DataSet,
     MetasType,
 )
@@ -23,6 +22,10 @@ from dataquality.utils.cv import (
     _img_path_to_b64_str,
     _img_to_b64_str,
 )
+
+# smaller than ITER_CHUNK_SIZE from base_data_logger because very large chunks
+# containing image data often won't fit in memory
+ITER_CHUNK_SIZE_IMAGES = 10000
 
 
 class ImageClassificationDataLogger(TextClassificationDataLogger):
@@ -141,7 +144,7 @@ class ImageClassificationDataLogger(TextClassificationDataLogger):
         *,
         imgs_colname: Optional[str] = None,
         imgs_location_colname: Optional[str] = None,
-        batch_size: int = ITER_CHUNK_SIZE,
+        batch_size: int = ITER_CHUNK_SIZE_IMAGES,
         id: Union[str, int] = "id",
         label: Union[str, int] = "label",
         split: Optional[Split] = None,
--- a/dataquality/loggers/data_logger/image_classification.py
+++ b/dataquality/loggers/data_logger/image_classification.py
@@ -3,7 +3,7 @@ from enum import Enum, unique
 from typing import List, Optional, Union
 
 import pandas as pd
-from PIL import Image
+from PIL.Image import Image
 
 from dataquality.exceptions import GalileoException
 from dataquality.loggers.data_logger.base_data_logger import (
@@ -19,7 +19,7 @@ from dataquality.loggers.logger_config.image_classification import (
     image_classification_logger_config,
 )
 from dataquality.schemas.split import Split
-from dataquality.utils.cv import _img_path_to_b64_str, _img_to_b64_str
+from dataquality.utils.cv import _img_path_to_b64_str, _img_to_b64_str, _bytes_to_b64_str
 
 
 @unique
@@ -52,15 +52,7 @@ class ImageClassificationDataLogger(TextClassificationDataLogger):
             inference_name=inference_name,
         )
 
-    @staticmethod
-    def _infer_image_field_type(example: Union[Image, dict, str]) -> ImageFieldType:
-        if isinstance(example, Image):
-            return ImageFieldType.pil_image
-        if isinstance(example, str):
-            return ImageFieldType.file_path
-        if isinstance(example, dict) and all(k in example for k in ["bytes", "path"]):
-            return ImageFieldType.hf_image_feature
-        return ImageFieldType.unknown
+
 
     def log_image_dataset(
         self,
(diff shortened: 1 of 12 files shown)


dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+        if imgs_colname is None and imgs_location_colname is None:
+            raise GalileoException(
+                "Must provide one of imgs_colname or imgs_location_colname."
+            )
+        if self.is_hf_dataset(dataset):
+            dataset = self._prepare_hf(dataset,
+                                       imgs_colname=imgs_colname,
+                                       imgs_location_colname=imgs_location_colname,
+                                       id_=id)
+        elif isinstance(dataset, pd.DataFrame):
```
Ben-Epstein, line 164: can we swap these? We created the `is_hf_dataset` so we can avoid importing the `datasets` library for as long as possible, because it can take a few seconds (!!) to load. 

If they pass in a pandas df, we shouldn't need to import datasets

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+            dataset = self._prepare_pandas(dataset,
+                                           imgs_colname=imgs_colname,
+                                           imgs_location_colname=imgs_location_colname,
+                                           imgs_dir=imgs_dir)
+        else:
+            raise GalileoException(
+                f"Dataset must be one of pandas or HF, " f"but got {type(dataset)}"
```
Ben-Epstein, line 171:
```suggestion
                f"Dataset must be one of pandas or HF, but got {type(dataset)}"
```
```suggestion
                f"Dataset must be one of pandas or HF, " f"but got {type(dataset)}"
```

Ben-Epstein, line 171:
```suggestion
                f"Dataset must be one of pandas or HF, but got {type(dataset)}"
```

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
         )
 
+    def _prepare_pandas(self,
+                        dataset: pd.DataFrame,
+                        imgs_location_colname: Optional[str],
+                        imgs_colname: Optional[str],
+                        imgs_dir: Optional[str],
+                        ) -> pd.DataFrame:
+        imgs_dir = imgs_dir or ""
+        imgs_dir: str
```
Ben-Epstein, line 60:
```suggestion
        imgs_dir: str = imgs_dir or ""
```

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+
+        # Find the id column, or create it.
+        if id_ not in dataset.column_names:
+            dataset = dataset.add_column(
+                name=id_, column=list(range(len(dataset)))
+            )
+
+        if imgs_colname is not None:
+            # HF datasets Image feature
+            import datasets
```
Ben-Epstein, line 98: you dont need this here, you import it at the top of the function

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+
+            if dataset.features[imgs_colname].dtype != "PIL.Image.Image":
```
Ben-Epstein, line 100: is there a way to do this comparison without comparing to a string? Can you use `isinstance` or actually the `from PIL.Image import Image` and compare that directly?

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+        import datasets
+        dataset: datasets.Dataset
+
+        # Find the id column, or create it.
+        if id_ not in dataset.column_names:
+            dataset = dataset.add_column(
+                name=id_, column=list(range(len(dataset)))
+            )
+
+        if imgs_colname is not None:
```
Ben-Epstein, line 96: I think the content inside of this `if` can be moved into a util or helper function in `dataquality.utils.image_classification` or something similar. The function is a bit long and harder to read

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -37,21 +50,126 @@ def __init__(
...
+                else:
+                    example["text"] = _bytes_to_b64_str(
+                        # assume abs paths for HF
+                        img_bytes=image["bytes"],
+                        img_path=image["path"],
+                    )
+                return example
+
+            dataset = dataset.map(hf_map_image_feature)
+        else:
```
Ben-Epstein, line 130: similarly here, I think breaking this out into a `process_hf_image_file_path` helper function or something similar would make this a lot easier to read

Ben-Epstein: Alternatively, if you can just break out that `hf_map_file_path` inner function into a top level helper, and call it from here, that would also make this code a lot more readable

Ben-Epstein (commented):
Awesome stuff @rfriel ! Fantastic PR description.

I think the `_prepare_hf` function can be decomposed a bit into some helper functions.

Also, as per your PR description, I think adding those other datasets (maybe some miniture versions for speed) would be great in terms of coverage here

dataquality/loggers/data_logger/image_classification.py
rfriel: Thanks!

dataquality/loggers/data_logger/image_classification.py
rfriel: I've changed to `isinstance(dataset.features[imgs_colname], datasets.Image)`, which is [effectively what the existing line did](https://github.com/huggingface/datasets/blob/2.8.0/src/datasets/features/image.py#L64).  Can't remember why I checked it in such a weird way originally

dataquality/loggers/data_logger/image_classification.py
rfriel: > Awesome stuff @rfriel ! Fantastic PR description.
> 
> I think the `_prepare_hf` function can be decomposed a bit into some helper functions.
> 
> Also, as per your PR description, I think adding those other datasets (maybe some miniture versions for speed) would be great in terms of coverage here
> 
> (it also looks like you need to run a `./scripts/format.sh` here)

Thanks!

I moved the `_prepare_hf` if/else steps into functions in another file, as suggested.  And added tests for MNIST and CIFAR-10.  Let me know what you think.

pyproject.toml
```diff
@@ -26,7 +26,7 @@ requires = [
     "cachetools==5.2.0",
     "importlib-metadata==4.12.0",
     "evaluate",
-    "datasets==2.6.1",
+    "datasets==2.6.2",
```
Ben-Epstein, line 29: i would suggest being less strict with this pin

```suggestion
    "datasets>=2.6",
```

Ben-Epstein (approved):
awesome! I'd loosen the pin on `datasets` but left a ✅

anthonycorletti (approved):
awesome stuff @rfriel ! what kinds of changes will we have to make to our colabs. well have some work to do when creating cv example notebooks.

rfriel:
> awesome stuff @rfriel ! what kinds of changes will we have to make to our colabs? well have some work to do when creating cv example notebooks.

I can look over one of the existing ones (say MNIST) and try to port it over.

The food101 one uses this code already, but it's kind of a pain to use as a test case, among other things because it's a big dataset.

PR #495, opened 2023-01-25
Upgrade datasets

rfriel:
Upgrades HF datasets dependency ~~to 2.6.2~~ (update: unpinned) to prevent the error

```python
TypeError: can only concatenate str (not "int") to str
```

that we've been seeing when trying to load some datasets.  See https://github.com/huggingface/datasets/issues/5406

Verified that `load_dataset('mnist')` works on 2.6.2, but not on 2.6.1.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ requires = [
     "cachetools==5.2.0",
     "importlib-metadata==4.12.0",
     "evaluate",
-    "datasets==2.6.2",
+    "datasets>=2.6",
     "transformers<4.25.0",
     "seqeval",
     "sentence-transformers>=2.2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ requires = [
     "cachetools==5.2.0",
     "importlib-metadata==4.12.0",
     "evaluate",
-    "datasets==2.6.1",
+    "datasets==2.6.2",
     "transformers<4.25.0",
     "seqeval",
     "sentence-transformers>=2.2",


anthonycorletti (approved):

Ben-Epstein (approved):

pyproject.toml
```diff
@@ -26,7 +26,7 @@ requires = [
     "cachetools==5.2.0",
     "importlib-metadata==4.12.0",
     "evaluate",
-    "datasets==2.6.1",
+    "datasets==2.6.2",
```
Ben-Epstein, line 29: we dont need to pin datasets so tightly here..
this should be fine 
```suggestion
    "datasets>=2.6",
```

rfriel: Sure, let's try it.

In general we ought to move toward much, much less pinning.  Right now, it's easy to imagine training setups that would break when dq is installed.

Ben-Epstein: yup i definitely agree

bogdan-galileo:
@rfriel Didn't we pin to 2.6.1 because the next one, 2.6.2 had issues ? Wasn't there a problem with the class `Accuracy` or something like this ?

rfriel:
> @rfriel Didn't we pin to 2.6.1 because the next one, 2.6.2 had issues ? Wasn't there a problem with the class `Accuracy` or something like this ?

The `Accuracy` issue was with `torchmetrics`.

`datasets` 2.6.2 is just 2.6.1 [plus a single change to fix this one issue](https://github.com/huggingface/datasets/compare/2.6.1...2.6.2).

PR #496, opened 2023-01-26
Setting unpatch on start to false per default

franz101:
- Try catch reference errors
- Unpatch on start to false per default
- Added better description to classifier_layer


Ben-Epstein (approved):

anthonycorletti (approved):
🙌 thanks @franz101

PR #499, opened 2023-02-06
XGB type checking only

elboy3:
I'm sure we will fully replace `: xgb.XGBClassifier` with sklearn estimator some time very soon, but just not yet. bigger fish to fry in short term


Ben-Epstein (approved):
Das smart

PR #502, opened 2023-02-08
Add features df and pred to data

elboy3:
Adding a `features.hdf5` DF and predictions to the data df

Produced on local 👇 
Screen Shot 2023-02-08 at 10 27 39 AM

### DIFF
--- a/dataquality/loggers/data_logger/structured_classification.py
+++ b/dataquality/loggers/data_logger/structured_classification.py
@@ -58,10 +58,13 @@ class StructuredClassificationDataLogger(BaseGalileoDataLogger):
             - If X is a numpy array, the feature names are provided
             - If X is a numpy array, the number of features in X and feature names
                 are the same
+            - Feature names match the feature names logged in a prior split
 
         Sets:
             - self.X to a pandas DataFrame if it is a numpy array
             - self.y to a numpy array if it is a list
+            - self.feature_names to the column names of X if it is a pandas DataFrame
+            - self.logger_config.feature_names to the column names of X if they aren't set
         """
         self.validate()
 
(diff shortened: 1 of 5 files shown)


dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -116,58 +116,70 @@ def log(self) -> None:
...
         )
+        assert self.probs is not None, "probs must be set before getting dfs"
+
         ids = self.X.id.values if "id" in self.X.columns else np.arange(len(self.X))
         n_rows = len(self.X)
 
-        df = vaex.from_pandas(self.X)
-        df["id"] = ids
+        feature_df = vaex.from_pandas(self.X)
+        feature_df["id"] = ids
```
Ben-Epstein, line 164: will `ids` ever be a list? If you're setting this in vaex, itll have to be a numpy array. So you may wanna do 
```suggestion
        feature_df["id"] = np.array(ids)
```
if it could be a list

dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -116,58 +116,70 @@ def log(self) -> None:
...
-        else:
-            probs_df["gold"] = self.y
+            df["inference_name"] = np.array([self.inference_name] * n_rows)
+
+        prob_df = vaex.from_arrays(
+            id=ids,
+            prob=self.probs,
+        )
+        if self.split != Split.inference:
+            prob_df["gold"] = self.y
```
Ben-Epstein, line 180: same here
```suggestion
            prob_df["gold"] = np.array(self.y)
```

Ben-Epstein (approved):

dataquality/loggers/data_logger/structured_classification.py
elboy3: it will always be numpy!

dataquality/loggers/data_logger/structured_classification.py
elboy3: same here. we ensure it in validation

PR #505, opened 2023-02-10
make the NOT case first

Ben-Epstein:
not even sure if this is a real thing in NER but i noticed we have the binary case backwards


anthonycorletti (approved):

elboy3 (approved):

Ben-Epstein:
on second thought this PR actually makes 0 sense. You can't have a "binary' case in NER, because you have a tagging schema that includes the B-, I-, and O tags (or others for different schemas). This line of code doesn't really mean anything...

PR #506, opened 2023-02-11
Fix NER HF token alignment for BILOU data

rfriel:

In `dataquality.utils.hf_tokenizer.LabelTokenizer`, there were a few accidental `if/if/else`s that should have been `if/elif/else`.

If the labels are BILOU, this causes the alignment process to convert them into BIO (unintended) while aligning them with HF tokens (intended).

This results in the model being trained to produce BIO spans, which the rest of dq will try to interpret as BILOU spans.  No valid BIO span is a valid BILOU span, so we'll record no predicted spans from the model ever.


rfriel:
Dunno what the failing test is about.  It doesn't import the code I changed.

Ben-Epstein (approved):

Ben-Epstein:
@rfriel this is a flaky test - i'm going to skip it and make a ticket for @elboy3 to check out

Ben-Epstein:
@elboy3 https://app.shortcut.com/galileo/story/3721

anthonycorletti (approved):

PR #508, opened 2023-02-13
Loose  versioning

franz101:
Make minimum requirements standard instead of ==
Fix only the ones needed.
Further I use Google Colab (pip freeze) as minimum version (to speed up installation).


tests/test_utils/spacy_integration.py
```diff
@@ -34,20 +35,20 @@ def train_model(
 
     # Galileo code
     watch(nlp)
-    log_input_examples(training_examples, "training")
-    log_input_examples(training_examples, "test")
+    log_input_examples(training_examples, Split.training)
+    log_input_examples(training_examples, Split.test)
```
elboy3, line 39: small detail but i think these were intentionally strings since they are tests (fine and even encouraged to be very manual / explicit in tests)

dataquality/utils/imports.py
```diff
@@ -23,3 +28,46 @@ def tf_available() -> bool:
...
+    import to help facilitate resolving relative imports.
+
+    to_import is an iterable of the modules to be potentially imported (absolute
+    or relative). The `as` form of importing is also supported,
+    e.g. `pkg.mod as spam`.
+
+    This function returns a tuple of two items. The first is the importer
+    module for easy reference within itself. The second item is a callable to be
+    set to `__getattr__`.
+    """
```
elboy3, line 46: Where is this coming from? can you link a stackoverflow link or blog if you are taking from that?

elboy3 (approved):

Ben-Epstein (approved):
very smart on colab!

tests/test_utils/spacy_integration.py
franz101: latest Mypy disagrees :)

dataquality/utils/imports.py
franz101: removed

PR #509, opened 2023-02-13
Fix save feature importances

elboy3:
We were running into issues where:

- if you overwrote a run name, it would delete the feature importances. This was because we wrote the feature importances before `finish()` so it would delete the run in the DB and cascade the metrics. We now save the feature importances in the data logger `upload` after resetting the run name, so we are fine.
- There was also an issue with not json serializing numpy floats, which is fixed with `tolist()`

These have been tested on this run
https://console.dev.rungalileo.io/insights?dataframeColumns=confidence%3B%26data_error_potential%3B%26gold%3B%26petal+length+cm%3B%26petal+width+cm%3B%26pred%3B%26sepal+length+cm%3B%26sepal+width+cm%3B%26text&projectId=ee785245-7d39-47f8-9f31-4cbf5023fb1a&runId=56175074-567e-47db-b578-3f59e0b72e90&taskType=4

With swagger 

Screen Shot 2023-02-13 at 3 51 41 PM

### DIFF
--- a/dataquality/loggers/data_logger/structured_classification.py
+++ b/dataquality/loggers/data_logger/structured_classification.py
@@ -22,6 +22,7 @@ from vaex.dataframe import DataFrame
 from dataquality.clients.objectstore import ObjectStore
 from dataquality.loggers.data_logger.base_data_logger import BaseGalileoDataLogger
 from dataquality.loggers.logger_config.structured_classification import (
+    StructuredClassificationLoggerConfig,
     structured_classification_logger_config,
 )
 from dataquality.schemas import __data_schema_version__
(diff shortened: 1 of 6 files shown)


dataquality/loggers/data_logger/structured_classification.py
```diff
@@ -144,11 +155,9 @@ def save_feature_importances(self) -> None:
 
         Assumes the model is fit
         """
-        suffix = "must be set before saving feature importances"
-        assert self.model, f"Model {suffix}"
-        assert self.feature_names, f"Feature names {suffix}"
-        assert config.current_project_id, f"Project ID {suffix}"
-        assert config.current_run_id, f"Run ID {suffix}"
+        template = " {inp} must be set before saving feature importances"
```
Ben-Epstein, line 158: noise

Ben-Epstein (approved):
i'd suggest not checking in the jupyter notebook every time, but it's all good for now

elboy3:
> i'd suggest not checking in the jupyter notebook every time, but it's all good for now

I made changes to the notebook - it would have been broken otherwise

PR #510, opened 2023-02-14
feat: auto_notebook

anthonycorletti:

No issue!

https://urshr-demo.notebook.us-west-1.sagemaker.aws/notebooks/galileo-demo.ipynb

***Demonstration***

See above!

🚀

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -141,6 +141,7 @@ __all__ = [
     "enable_galileo_verbose",
     "enable_galileo",
     "auto",
+    "auto_notebook",
 ]
 
 try:
--- a/dataquality/notebook.py
+++ b/dataquality/notebook.py
@@ -1,0 +1,123 @@
+from datetime import datetime
+from functools import partial
+from typing import Any, Optional
+
+TRAIN_CSV = "training.csv"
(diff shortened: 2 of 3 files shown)


Ben-Epstein (commented):
@anthonycorletti i thought that the requirement involved having an upload button? The "Run Galileo" button is great but it assumed the user already uploaded their data. 

Happy to help with that upload button if you'd like (if we need it), lmk

anthonycorletti:
> @anthonycorletti i thought that the requirement involved having an upload button? The "Run Galileo" button is great but it assumed the user already uploaded their data.
> 
> Happy to help with that upload button if you'd like (if we need it), lmk

good point! so the upload file buttons turned into more work than was worth in a jupyter environment, atleast with the approach i was taking, (jupyter upload file extension). it required nodejs in the notebook env via conda-forge and i wasnt waiting around for that to install, and neither should anyone else lol. if youd like to try another approach go for it!

for uploading files we could simply educate the user on how to navigate to `/tree` in a notebook environment and use the file upload functionality there to do the same thing. worth noting, this is a stopgap solution for eventual ui support.

anthonycorletti:
closing wrt to #516

PR #515, opened 2023-02-15
Sphinx documentation

franz101:
Allows do to create autodocs :)


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Pushes to another repository
+        uses: cpina/github-action-push-to-another-repository@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.GH_TOKEN }}
+        with:
+          source-directory: 'docs/autodocs/_build/markdown'
+          target-directory: 'autodocs'
+          destination-github-username: 'rungalileo'
+          destination-repository-name: 'docs'
+          user-email: franz@rungalileo.io
```
elboy3, line 38: should we use a shared email like galileo@rungalileo.io?

elboy3 (approved):

.github/workflows/docs.yaml
Ben-Epstein, line 38:
```suggestion
          user-email: team@rungalileo.io
```

PR #526, opened 2023-02-23
FastAI callback

franz101:
The PR for fastai got good feedback, but didn't get merged, therefore the code is ported to our module:
https://github.com/fastai/fastai/pull/3874

There is an end to end test notebook and currently a client is relying on FastAI, therefore we should push this out this week.


pyproject.toml
```diff
@@ -66,6 +66,7 @@ test = [
     "torchtext>=0.13.1",
     "torchdata>=0.4.1",
     "xgboost>=1.6.2",
+    "timm"
```
anthonycorletti, line 69: @franz101 should we pin this?

dataquality/integrations/fastai.py
```diff
@@ -0,0 +1,370 @@
...
+        :param x_col: Name of the column to use for the x values, for example text.
+        :param y_col: Name of the column to use for the y values, for example label.
+        :return: Pandas DataFrame with the data from the DataLoader.
+        """
+        df = dl.items.copy()
+        rename_options = {}
+        if "x_col" != "text":
+            rename_options[x_col] = "text"
+        if "y_col" != "text":
+            rename_options[x_col] = "label"
```
anthonycorletti, line 365: should this be 

```
if "y_col" != "text":
            rename_options[x_col] = "label"
```

?

anthonycorletti (approved):
non blocking comments inline!

dataquality/integrations/fastai.py
```diff
@@ -0,0 +1,370 @@
...
+        of the dataset. This helps to find mislabelled samples in a
+        data centric approach.
+        :param layer: Classifier layer with embeddings as input and logits as output.
+        :param log_dataset: Enable automatic extraction of the dataset to data quality.
+        :param disable_dq: Disable data quality logging.
+        :param args: The arguments to pass to the super class.
+        :param kwargs: The keyword arguments to pass to the super class.
+        """
+        super().__init__(*args, **kwargs)
+        self.disable_dq = os.environ.get("DQ_NOOP", False)
```
Ben-Epstein, line 111: we have a global helper for this already, please use that https://github.com/rungalileo/dataquality/blob/main/dataquality/utils/helpers.py#L30

```suggestion
        self.disable_dq = galileo_disabled()
```

dataquality/integrations/fastai.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def __init__(
+        self,
+        labels: List[str],
+        layer: Any = None,
+        log_dataset: bool = True,
+        task_type: str = "image_classification",
+        options: Dict[str, Any] = {},
+        *args: Any,
+        **kwargs: Any,
```
Ben-Epstein, line 98: how will people know that they can provide a project and run name?

dataquality/integrations/fastai.py
franz101: good catch

dataquality/integrations/fastai.py
franz101: It's in the docs at the moment but maybe something like pedantic for each watch function, would be the best solution. what do you think?

dataquality/integrations/fastai.py
```diff
@@ -0,0 +1,370 @@
...
+                return
+        except Exception as e:
+            print("cur_split error", cur_split, e)
+            return
+        # Log the model outputs
+        embs = self.model_outputs_log[FastAiKeys.model_input][0].detach().cpu().numpy()
+        logits = self.model_outputs_log[FastAiKeys.model_output].detach().cpu().numpy()
+        equal_len = len(embs) == len(logits) and len(ids) == len(embs)
+        if not equal_len:
+            print(
```
Ben-Epstein, line 290: shouldn't we raise an exception here? OTherwise it's going to call `log_model_outputs` and raise anyway

dataquality/integrations/fastai.py
Ben-Epstein: or perhaps actually 
```
if "y_col" != "label":
    rename_options[x_col] = "label"
```

pyproject.toml
Ben-Epstein: +1 let's lower pin at least `>=X`

dataquality/integrations/fastai.py
```diff
@@ -0,0 +1,370 @@
...
+                dataquality.log_image_dataset(
+                    self.convert_img_dl_to_df(valid_dl),
+                    imgs_colname="image",
+                    imgs_location_colname="path",
+                    split=dataquality.schemas.split.Split.validation,
+                )
+        else:
+            if train_dl is not None:
+                dataquality.log_dataset(
+                    self.convert_tab_dl_to_df(train_dl),
```
Ben-Epstein, line 209: based on the way that `convert_tab_dl_to_df` works at the moment, it means that we only support image classification and text classification for FastAI. Please add a validation at the top that checks the task_type and immediately raises if it's not one of those

dataquality/integrations/fastai.py
Ben-Epstein: pydantic is nice for sure! But in the meantime can you just add those 2 as optionals?
```
    def __init__(
        self,
        labels: List[str],
        layer: Any = None,
        log_dataset: bool = True,
        task_type: str = "image_classification",
        options: Dict[str, Any] = {},
        project_name: Optional[str] = None,
        run_name: Optional[str] = None,
        *args: Any,
        **kwargs: Any,
```

dataquality/integrations/fastai.py
Ben-Epstein: wait @franz101 i think it needs to be what I wrote right?

Ben-Epstein (changes requested):
looks great! a few small changes

dataquality/integrations/fastai.py
franz101: removed for now

PR #536, opened 2023-03-14
Fix/keras watch

bogdan-galileo:
Fix issue in the keras watch function discovered while helping Kendal from MC debug her Keras notebook.

Allow `layer` to be None (adding the optional type is not enough) + fix if/else logic.

### DIFF
--- a/dataquality/integrations/experimental/keras.py
+++ b/dataquality/integrations/experimental/keras.py
@@ -280,14 +280,16 @@ def store_model_ids(store: Dict[str, Any]) -> Callable:
 
 
 def select_model_layer(
-    model: tf.keras.layers.Layer, layer: Optional[Union[tf.keras.layers.Layer, str]]
+    model: tf.keras.layers.Layer, layer: Union[tf.keras.layers.Layer, str] = None
 ) -> tf.keras.layers.Layer:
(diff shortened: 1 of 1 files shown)


dataquality/integrations/experimental/keras.py
```diff
@@ -280,14 +280,16 @@ def train_step_wrapper(orig_func: Callable, *args: Any, **kwargs: Any) -> None:
 
 
 def select_model_layer(
-    model: tf.keras.layers.Layer, layer: Optional[Union[tf.keras.layers.Layer, str]]
+    model: tf.keras.layers.Layer, layer: Union[tf.keras.layers.Layer, str] = None
```
Ben-Epstein, line 283:
```suggestion
    model: tf.keras.layers.Layer, layer: Union[tf.keras.layers.Layer, str, None] = None
```

Ben-Epstein: need to still indicate in typing that it can be none

bogdan-galileo: Oh, I guess that was the point of the `Optional` and could have left it. Thanks

elboy3 (approved):

PR #537, opened 2023-03-17
filter and warn on empty tokens

Ben-Epstein:
@rfriel @jonathangomesselman 

image


rfriel (approved):
:) 

Should we add this to the spacy NER data logging path too?

Ben-Epstein:
> :)
> 
> Should we add this to the spacy NER data logging path too?

I kind of expect that Spacy will handle this during their own validation on creating the input `Examples`

PR #541, opened 2023-03-21
Chore/small cleanups

Ben-Epstein:
small improvements here and there

1. fix the `configure` warning when calling `set_console_url` - add an `_internal` flag that removes warning
2. downcast to float32 when logging. This saves huge amounts of space if they are logging f64 (like i was today during testing amazon polarity)
3. in our hf tokenizer, first filter then add IDs, just to make the ids continuous (noop, just nicer)
4. add a print to encourage people to install dq with the cuda extra
5. remove the deprecation on probs, users can use whatever they prefer
6. Do incremental PCA instead of PCA during the GPU UMAP phase, because it's more memory efficient
7. use `with torch.autocast("cuda"):` when creating data embeddings, it speeds things up by around 10 sentences/sec during testing with amazon polarity (thanks Derek!)

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -30,7 +30,7 @@ If you want to train without a model, you can use the auto framework:
         dataquality.get_insights()
 """
 
-__version__ = "v0.8.21"
+__version__ = "v0.8.23"
 
 import sys
 from typing import Any, List, Optional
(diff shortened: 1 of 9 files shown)


rfriel (approved):
LGTM, thanks!

PR #549, opened 2023-03-29
fix: [sc-4518] exoscale presigned url rework

anthonycorletti:
Story details: https://app.shortcut.com/galileo/story/4518

### DIFF
--- a/dataquality/clients/objectstore.py
+++ b/dataquality/clients/objectstore.py
@@ -5,6 +5,7 @@ from tempfile import NamedTemporaryFile
 from typing import Optional
 
 import requests
+from minio import Minio
 from tqdm.auto import tqdm
 from tqdm.utils import CallbackIOWrapper
 from vaex.dataframe import DataFrame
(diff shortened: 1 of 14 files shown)


dataquality/clients/objectstore.py
```diff
@@ -13,10 +14,32 @@
...
+                "environment variables must be set. "
+                "Please set these variables with the values from your Exoscale "
+                "API Key."
+            )
+            self._minio_client = Minio(
+                endpoint=config.minio_fqdn,
+                access_key=access_key,
+                secret_key=secret_key,
+                secure=True,
+            )
```
Ben-Epstein, line 41: can we move this into a helper like `self._handle_exoscale` where we can have a docstring explaining why this is necessary for exoscale?

dataquality/clients/objectstore.py
```diff
@@ -32,14 +55,42 @@ def create_object(
...
+                progress=progress,
+            )
+
+    def _create_object_exoscale(
+        self,
+        object_name: str,
+        file_path: str,
+        content_type: str = "application/octet-stream",
+        bucket_name: Optional[str] = None,
+    ) -> None:
```
Ben-Epstein, line 87: docstring

pyproject.toml
```diff
@@ -33,7 +33,8 @@ requires = [
     "Pillow",
     "h5py >=3.1.0",
     "numpy<1.24.0",
-    "tenacity>=8.1.0"
+    "tenacity>=8.1.0",
+    "minio>=7.1.0,<7.2.0",
```
Ben-Epstein, line 37: thoughts on making an `exoscale` extra? So we can do `pip install dataquality[exoscale]` instead of needing it globally for the dq library?

anthonycorletti: good point. thinking pip install dataquality[minio] for now so we can tack it on for other kinds of installs that might just need the minio package.

Ben-Epstein (approved):
amazing!!

PR #555, opened 2023-04-10
quickfix

Ben-Epstein:
bug during parse label. I made it more robust


dataquality/loggers/data_logger/text_classification.py
```diff
@@ -1,12 +1,13 @@
...
 from enum import Enum, unique
-from typing import Any, DefaultDict, Dict, Iterable, List, Optional, Union
+from typing import Any, DefaultDict, Dict, Iterable, List, Optional, Union, cast
 
 import numpy as np
 import pandas as pd
 import vaex
 from vaex.dataframe import DataFrame
 
+import dataquality
```
setu4993, line 10: Nit: Might be nicer to import the function directly to match other imports below.
```suggestion
from dataquality import set_labels_for_run
```

dataquality/loggers/data_logger/text_classification.py
```diff
@@ -266,16 +267,27 @@ def _log_hf_dataset(
...
-            try:
-                parse_label = lambda x: dataset.features[label].int2str(x)  # noqa: E731
-            except Exception:
+        def parse_label(labels: Union[List[int], List[str]]) -> List[str]:
+            # If we have 1 str, they are all strings
+            if isinstance(labels[0], str):
+                return cast(List[str], labels)
+            # Otherwise they are all ints (typing)
+            else:
+                labels = cast(List[int], labels)
```
setu4993, line 276: Should we be checking here if they are actually ints?

setu4993 (commented):
Left a tiny note while I'm continuing to review the tests.

tests/loggers/test_text_classification.py
```diff
@@ -367,3 +368,22 @@ def test_create_and_upload_data_embs(
...
+        mock_method.return_value = logger
+        dq.log_dataset(TRAIN_HF_DS, split="train")
+        # We should grab the labels from the dataset and save them
+        assert logger.logger_config.labels == TC_LABELS
+
+    new_logger = TextClassificationDataLogger()
+    with mock.patch("dataquality.core.log.get_data_logger") as mock_method:
+        mock_method.return_value = new_logger
+        # Even thought the test ds doesn't have the label names, we should be able
+        # to grab them from the config.labels because we grabbed them from the train ds
```
setu4993, line 388: Nice.

tests/test_utils/tc_constants.py
```diff
@@ -0,0 +1,137 @@
+from datasets import Dataset
+
+TC_JSON_DATA = [
```
setu4993, line 3:
```suggestion
TEXT_CLASSIFICATION_JSON_DATA = [
```

tests/test_utils/tc_constants.py
```diff
@@ -0,0 +1,137 @@
...
+    "Cancel Payment",
+    "Card Issue",
+    "Card Lost_and_Card Replacement",
+    "Card Not Received",
+    "Customer Insult",
+    "FAQ",
+    "Interest charge",
+    "Mailing status",
+    "Report Fraud",
+    "See Activity",
```
setu4993, line 127: IMHO would definitely be nicer to have a dict here instead of a list.

setu4993 (approved):
Nice fix, Ben! But I'm also only just getting familiar with all that's going on here, in case you want to wait for someone else's review.

tests/test_utils/tc_constants.py
Ben-Epstein: a dict of what?

dataquality/loggers/data_logger/text_classification.py
Ben-Epstein: they would only ever be strings or ints, and it wouldn't really make sense for there to be a mix of strings and ints in your training dataset

tests/test_utils/tc_constants.py
setu4993: Dict of ints (from `TC_JSON_DATA`) -> str (`TC_LABELS`).

dataquality/loggers/data_logger/text_classification.py
setu4993: Sure, but as I understand it, we can't control the upstream datasets, so it's worth checking if it is actually what we expect it to be.

tests/test_utils/tc_constants.py
Ben-Epstein: I'm not sure what you mean exactly - we want to pass in the huggingface `Dataset` directly into the function, and test that we can grab the `name` attribute

setu4993: It's possible I'm missing something... I was trying to get at avoiding the duplicative maps on L132 and L137.

I think I misunderstood those, though, never mind.

Ben-Epstein: oh I see. Those are adding an `id` column to the dataset, not mapping the labels. I definitely could have added it to the `TC_JSON_DATA` as another column, but I just took this code from the customer bug report, and I think the maps are OK since it's just 100 samples, probably insignificant time to run at test startup

dataquality/loggers/data_logger/text_classification.py
Ben-Epstein: Yea i hear that, but I'm not sure it's worth the processing time for such an unlikely edge case. The odds that a user has a dataset with a mix of label names and label indices in their dataset seems almost impossible (you couldn't train your model like that anyway). So in this case I do think checking just the first value should be sufficient

anthonycorletti (approved):

PR #563, opened 2023-04-14
Fixes/memory leak img

franz101:
Temporary fix that chunks the image files upload without loading all images into memory.
Caveat. I'm not sure if the file is still uploaded twice as I can see a file exists check and upload the blob data,


Ben-Epstein (approved):
im gonna ✅  this because it's blocking a customer, but i'd like to chat through this a bit and see if there are any better implementations available

anthonycorletti (approved):

PR #567, opened 2023-04-20
DQYOLO optimization

franz101:
- Move validation to beginning
- Add more explanations
- Added further validation
- Added iou and nms from parameters


dataquality/utils/dqyolo.py
```diff
@@ -0,0 +1,113 @@
...
+
+def get_iou_thres(arguments: list) -> float:
+    """Extract the iou threshold from the arguments of yolo.
+
+    :param arguments: The arguments of ultralytics yolo.
+    :return: The iou threshold.
+    """
+    value = extract_value(arguments, "iou")
+    if not value:
+        return 0.7
```
Ben-Epstein, line 55: should we make it a constant? And should we make it 0.5, since that's the default in the UI?

dataquality/dqyolo.py
```diff
@@ -80,6 +46,48 @@ def main() -> None:
...
+        tmp_cfg_path = cfg.get(ultralytics_split_mapping[split])
+        if not tmp_cfg_path:
+            continue
+        relative_img_paths[split] = cfg[f"bucket_{ultralytics_split_mapping[split]}"]
+    if not len(relative_img_paths):
+        raise ValueError("No dataset paths found in config file.")
+
+    # 2. Init galileo
+    project_name = (
+        extract_value(original_cmd, "project")
```
bogdan-galileo, line 71: Mmmh, isn't that one something that is used in Yolo ? I think I'd remove that, especially that it takes precedence over the environ var. Imagine the scenario where they hardcoded project=top_secret_research in all their yaml files and that overwrites all their runs in the console...

Also, once you're in a console to launch your `dqyolo` command, I feel like it's easier to set the environ variable rather than editing your YAML. No ?

dataquality/dqyolo.py
```diff
@@ -80,6 +46,48 @@ def main() -> None:
...
+    relative_img_paths: Dict[Split, str] = {}
+    # Check each file
+    for split in [Split.training, Split.validation, Split.test]:
+        # Create a temporary config file for the training set that changes
+        # the dataset path to the validation set so we can log the results
+        tmp_cfg_path = cfg.get(ultralytics_split_mapping[split])
+        if not tmp_cfg_path:
+            continue
+        relative_img_paths[split] = cfg[f"bucket_{ultralytics_split_mapping[split]}"]
+    if not len(relative_img_paths):
```
bogdan-galileo, line 66: I was thinking you even try to ping or retrieve one image to ensure that there is no issue with typos + the fact that the data might be unreachable, etc. Is that a lot more work ? We can add to feature request if yes

dataquality/dqyolo.py
franz101: I personally not a big fan of env variables because it makes the command longer and if things crash sometimes i need to set them again, would suggest to support both.

dataquality/dqyolo.py
franz101: Lets do that a feature request, this need to be done once the first batch is logged

Ben-Epstein (approved):

dataquality/dqyolo.py
bogdan-galileo: wait with env vars I was thinking you do `export GALILEO_PROJECT_NAME = cool_project` in your terminal once and even if the training loop or something crashes, you don't have to do it again, no ?

Yeah I like `dq_project` actually!

bogdan-galileo (approved):
Awesome

PR #568, opened 2023-04-24
Adjust pca num components for tiny datasets

bogdan-galileo:
`pca.partial_fit` breaks when called with a (small, test) dataset with less than `PCA_N_COMPONENTS=100` elements. This only happens in OD because that's the only place where this method is called (to avoid uploading huge embeddings, which have dimension 2048 for the DETR architecture)

Screenshot 2023-04-24 at 10 48 27 AM

Tested with run https://console.dev.rungalileo.io/insights?projectId=37b6f14d-f1c8-4164-8bef-4773b883f819&runId=448a5f45-42cd-47b4-9618-15caa133fd23&taskType=5


elboy3 (commented):
Just making sure I understand this right, since n_components is small we aren't worried about bringing all of the values into memory? 

Would be awesome if we can add some kind of small test as well that tests len(df) < PCA_N_COMPONENTS and makes a few assertions about the resulting emb_pca col

bogdan-galileo:
> Just making sure I understand this right, since n_components is small we aren't worried about bringing all of the values into memory?
> 
> Would be awesome if we can add some kind of small test as well that tests len(df) < PCA_N_COMPONENTS and makes a few assertions about the resulting emb_pca col

Yeah exactly, the max size it can have is 99x99 so it's fine to bring in to memory. Good point about tests, added!

tests/assets/constants.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+TEST_ASSETS_DIR = f"{os.getcwd()}/tests/assets"
```
elboy3, line 3: can we use numpy with a fixed random seed instead a .npy file?

tests/utils/test_vaex_utils.py
```diff
@@ -14,3 +17,17 @@ def test_drop_empty_cols() -> None:
...
+    # DETR returns embeddings of dimension 2048.
+    df = vaex.from_arrays(
+        x=[1, 2, 3], y=[4, 5, 6], z=[7, 8, 9], emb=np.random.rand(3, 2048)
+    )
+    df = add_pca_to_df(df)
+
+    pca_embs = df["emb_pca"].values
+    assert pca_embs.shape == (3, PCA_N_COMPONENTS)
+    assert np.any(pca_embs[:, :3])
+    assert not np.any(pca_embs[:, 3:])
```
elboy3, line 33: can you add 1-liner comments what these are asserting

tests/utils/test_vaex_utils.py
```diff
@@ -14,3 +17,17 @@ def test_drop_empty_cols() -> None:
...
+def test_add_pca_to_df() -> None:
+    np.random.seed(NUMPY_SEED)
+    # DETR returns embeddings of dimension 2048.
+    df = vaex.from_arrays(
+        x=[1, 2, 3], y=[4, 5, 6], z=[7, 8, 9], emb=np.random.rand(3, 2048)
+    )
+    df = add_pca_to_df(df)
+
+    pca_embs = df["emb_pca"].values
+    assert pca_embs.shape == (3, PCA_N_COMPONENTS)
```
elboy3, line 31: nit-pick but in the test can we hard-code the actual value of PCA_N_COMPONENTS? so put like 100 (if that's what it is). in tests hard code is actually good practice, helps you catch mistakes down the line if. /when vars change

bogdan-galileo: @elboy3 My thought was that if we change the value of PCA_N_COMPONENTS, then this test could become irrelevant (it tests a scenario that will never happen), no ?

Despite this, would you still hardcode PCA_N_COMPONENTS ?

dataquality/utils/vaex.py
```diff
@@ -214,15 +214,28 @@ def get_output_df(
...
         pca.partial_fit(chunk)
 
     @vaex.register_function()
     def apply_pca(emb: np.ndarray) -> np.ndarray:
         return pca.transform(emb).astype(np.float32)
 
-    df_copy["emb_pca"] = df_copy["emb"].apply_pca()
+    # pad the embeddings with zeros if the dimension is less than PCA_N_COMPONENTS
+    emb_values = df_copy["emb"].apply_pca()
+    if n_components < PCA_N_COMPONENTS:
```
Ben-Epstein, line 229: I'm not sure i understand this change just yet. What is this code doing? Why can't we just do <100 components as the target?

Ben-Epstein (commented):
1q

dataquality/utils/vaex.py
bogdan-galileo: I wasn't sure if there is a requirement down the road where it needs to be 100. I tried and it worked with less dimensions too, but since the difference between a, say, (10,10) array and (10,100) is minimal, I just padded for consistency.

Ben-Epstein:
@bogdan-galileo I'd prefer we go down to fewer components rather than padding. It'll reduce the complexity of the code a lot and will make it consistent with runners (which is what we do for other tasks when we apply PCA on the server)

Ben-Epstein (approved):

PR #574, opened 2023-05-04
added setfit support

franz101:
Due to recent demand, I have converted the notebook to support setfit in the following way:
```python
from dataquality.integrations.setfit import watch

dq.init(task_type=TaskType.text_classification)
labels = ["nocat", "cat"]
dq.set_labels_for_run(labels)
split = "training"
batch_size = 3
ds_len = len(dataset)
dq_evaluate = watch(model)
dataset = dataset.map(lambda x, idx: {"id": idx}, with_indices=True)
for i in range(0, ds_len, batch_size):
    print("batch", i)
    batch = dataset[i : i + batch_size]
    dq_evaluate(batch, split=split)

dq.finish()
```

[Issue in setfit repo](https://github.com/huggingface/setfit/issues/356)

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -30,7 +30,7 @@ If you want to train without a model, you can use the auto framework:
         dataquality.get_insights()
 """
 
-__version__ = "v0.8.37"
+__version__ = "v0.8.38"
 
 import sys
 from typing import Any, List, Optional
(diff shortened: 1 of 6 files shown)


franz101:
Here is the original implementation:
https://colab.research.google.com/drive/1hRDQ3Dafxd1Pr3Bz8pxEgbBldp9RAzvF?usp=sharing

Ben-Epstein (approved):

PR #578, opened 2023-05-05
Changed error param and removed early breaks

dcaustin33:
Small changes to rename variables to the correct names as we are always using ground truth polygons for error types. Also removed early iteration stopping in our watch code and instead just made the dataset smaller for better testing


anthonycorletti (approved):

dataquality/utils/semantic_segmentation/errors.py
```diff
@@ -42,15 +69,20 @@ def calculate_misclassified_polygons(
...
             predicted polygon map for one image
     """
-    for polygon in pred_polygons:
-        out_polygon = draw_polygon(polygon, gt_mask.shape[-2:])
-        if polygon_accuracy(gt_mask, out_polygon) < ERROR_THRES:
+    for polygon in gt_polygons:
+        out_polygon = draw_polygon(polygon, pred_mask.shape[-2:])
+        if polygon_accuracy(pred_mask, out_polygon) < ERROR_THRES:
             polygon.error_type = ErrorType.classification
+            polygon.misclassified_class_label = calculate_misclassified_class(
```
elboy3, line 76: rather than calling "polygon_accuracy" and checking error thres, and then doing it again in `calculate_misclassified_class` i think we should have all the logic for check error type and getting the misclassified class in the same fn

PR #579, opened 2023-05-05
Test splitting

franz101:
From 12 minutes back 5 minutes


franz101:
@setu4993 really need your help caching the installs.
Idea:
- We have a seperate action that caches the installs everday, then we use artifacts or a docker image to load the installs.
- Then we seperate the linting from the splitted tests.
This should get us back down to 6 minutes hopefully

.test_durations
```diff
@@ -0,0 +1,455 @@
+{
+    "tests/auto/test_ner_auto.py::test_call_auto_pandas_train_df": 0.013364318999990132,
+    "tests/auto/test_ner_auto.py::test_call_auto_pandas_with_inference": 0.01725827300001015,
```
elboy3, line 3: let's remove this file

scripts/pretest.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -ex
+
+. ./scripts/set-local-env.sh
```
elboy3, line 3: i don't think we want to set local env vars for ci testing

setu4993: Yeah, we shouldn't need any local vars for testing. And the ones we do, we should set them via `pytest-env` so it's not dependent on anyone using

.github/workflows/test.yaml
```diff
@@ -24,16 +25,43 @@ jobs:
...
             scripts/install.sh
           fi
+      - name: pretest
+        run: scripts/pretest.sh
       - name: test
-        run: scripts/test.sh
-      - name: upload coverage
-        uses: codecov/codecov-action@v1
+        run: |
+          pytest --cov=dataquality --cov=tests --cov-report=term-missing --cov-report=xml --durations=10 -o console_output_style=progress --splits 5 --group ${{ matrix.group }}
```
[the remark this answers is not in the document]
setu4993: I think we should continue to move away from `test.sh` into hosting config for our test suites that is native and agnostic to how it gets run.

That would be `pyproject.toml`.

setu4993 (commented):
Hey @franz101, a few thoughts for this one:

Let's avoid using a `pretest.sh` and move all of that config to a way that is more standard in the Python world. Running specific shell scripts before tests makes them harder to work with.

Separately, it'd be nicer to not split this into multiple groups and figure out a better way to parallelize the tests (`pytest-xdist`, maybe?) so that we avoid the overhead associated with maintaining the `.test_durations`.

setu4993:
> * We have a seperate action that caches the installs everday, then we use artifacts or a docker image to load the installs.

I think just using the plain GH Actions' `pip` cache setup is a good start. Let's see how much better it makes things.

> * Then we seperate the linting from the splitted tests.

Agreed! We started doing that for our internal repos and would highly advise moving lint / format steps out of test steps. Those have a separate set of dependencies (dev dependencies) that are unrelated to the primary project.

.github/workflows/test.yaml
```diff
@@ -24,16 +25,43 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: "pip"
```
Ben-Epstein, line 28: from https://github.com/rungalileo/rungalileo/blob/main/.github/workflows/test.yaml#LL48C11-L48C50

```suggestion
          cache: "pip"
          cache-dependency-path: "pyproject.toml"
```

.github/workflows/test.yaml
```diff
@@ -24,16 +25,43 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest pytest-cov pytest-split
+      - name: Download all artifacts
+        # Downloads coverage1, coverage2, etc.
+        uses: actions/download-artifact@v2
+      - name: Run coverage
+        run: |
+          coverage combine coverage*/.coverage*
+          coverage report --fail-under=100
```
Ben-Epstein, line 65:
```suggestion
          coverage report --fail-under=80
```

PR #581, opened 2023-05-08
Chore/url rename

dcaustin33:
Changed the url path to return a correct bucket and path that can be used in the ui


dataquality/integrations/cv/torch/semantic_segmentation.py
```diff
@@ -113,7 +118,7 @@ def convert_dataset(self, dataset: Any) -> List:
                 print("Only logging 100 images for now")
                 break
         # I have assumed we can collect the masks from the hooks in the dataloader
-        return processed_dataset
+        return processed_dataset, i + start_int + 1
```
Ben-Epstein, line 121: can you add a comment explaining `i + start_int + 1`

elboy3, line 121: we've already done `i += start_int` above, do we need to add start int again? 

why plus 1? 

mayve add to the docstring for convert_dataset what we are doing with start int

elboy3 (approved):

PR #583, opened 2023-05-09
Bug/handle hf down

Ben-Epstein:
seems like HF has gone down a few times (a customer complained, and an [IT failed](https://github.com/rungalileo/deploy/actions/runs/4925440125/jobs/8800893962)).

I am having issues as well
image

This quick fix will get us past this hf bug for now, because DQ shouldnt fail due to data embeddings issues

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -30,7 +30,7 @@ If you want to train without a model, you can use the auto framework:
         dataquality.get_insights()
 """
 
-__version__ = "v0.9.0"
+__version__ = "v0.8.35"
 
 import sys
 from typing import Any, List, Optional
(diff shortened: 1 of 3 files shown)


Ben-Epstein:
https://app.shortcut.com/galileo/story/5163

dataquality/__init__.py
```diff
@@ -30,7 +30,7 @@
         dataquality.get_insights()
 """
 
-__version__ = "v0.9.0"
+__version__ = "v0.8.35"
```
setu4993, line 33: Why are we going back versions?

setu4993 (commented):
Haven't reviewed the entire changeset but the version looks incorrect to me.

dataquality/__init__.py
Ben-Epstein: because it seems we aren't ready to deploy 0.9 yet. I'd just like to keep this in the 8's until we are all ready to bump up.

I'm going to cut a release after merging

setu4993: I see. Makes sense. Bumping up versions preemptively feels incorrect, but we can take that discussion elsewhere.

PR #586, opened 2023-05-10
Rename gt to gold for sem seg

elboy3:


setu4993 (approved):
Curious about the impetus for this change?

dcaustin33 (approved):

elboy3:
> Curious about the impetus for this change?

We use gold everywhere else in dq, never really made since to have gt for semseg

PR #588, opened 2023-05-11
add width and height to log

Ben-Epstein:
width and height are necessary for OD dataframes

relates to https://app.shortcut.com/galileo/story/5206/edits-cart-fix-coordinates-format

### DIFF
--- a/dataquality/integrations/ultralytics.py
+++ b/dataquality/integrations/ultralytics.py
@@ -370,6 +370,8 @@ class Callback:
                     ODCols.image: str(file_path),
                     ODCols.bbox: bbox_gold,
                     ODCols.gold_cls: image["cls"],
+                    ODCols.width: width,
+                    ODCols.height: height,
                 }
             )
         return processed_dataset
--- a/dataquality/integrations/ultralytics.py
+++ b/dataquality/integrations/ultralytics.py
@@ -348,10 +348,12 @@ class Callback:
             self.file_map[image["im_file"]] = i
             batch_img_shape = image["img"].shape[-2:]
             shape = image["ori_shape"]
+            image_width, image_height = shape
             bbox = image["bboxes"].clone()
-            height, width = batch_img_shape
+            # Batch width and height. The max width/height for this batch
+            bch_height, bch_width = batch_img_shape
             tbox = box_convert(bbox, "cxcywh", "xyxy") * torch.tensor(
-                (width, height, width, height), device=bbox.device
+                (bch_width, bch_height, bch_width, bch_height), device=bbox.device
             )
             ratio_pad = image["ratio_pad"]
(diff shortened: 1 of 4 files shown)


dataquality/loggers/data_logger/object_detection.py
```diff
@@ -83,13 +89,15 @@ def _get_input_df(self) -> DataFrame:
...
+            width=self.width,
+            height=self.height,
             **self.meta,
             # bbox=self.bbox,
             # cls=self.cls,
         )
         if self.split == Split.inference:
             inp["inference_name"] = [self.inference_name] * df_len
-
+        # import pdb; pdb.set_trace()
```
anthonycorletti, line 100: 🧹

anthonycorletti (approved):

dataquality/loggers/data_logger/object_detection.py
bogdan-galileo, line 100: remove this ?

anthonycorletti (approved):

dataquality/loggers/data_logger/object_detection.py
```diff
@@ -61,10 +63,14 @@ def __init__(
         split: Optional[str] = None,
         meta: Optional[MetasType] = None,
         inference_name: Optional[str] = None,
+        width: Optional[List[float]] = None,
+        height: Optional[List[float]] = None,
```
bogdan-galileo, line 67: You use `int` as their type below, should these match ?

dataquality/integrations/ultralytics.py
```diff
@@ -370,6 +370,8 @@ def convert_dataset(self, dataset: Any) -> List:
                     ODCols.image: str(file_path),
                     ODCols.bbox: bbox_gold,
                     ODCols.gold_cls: image["cls"],
+                    ODCols.width: width,
```
bogdan-galileo, line 373: I think we should set `width` and `height` from `shape = image["ori_shape"]` see a few lines above https://github.com/rungalileo/dataquality/blob/b04261ce265f2c4f613ced91e9d17f64c30fbe16/dataquality/integrations/ultralytics.py#LL350C13-L350C39.

The ones you're setting here are probably the w/h of the batch which I think is the max of any image in the batch

dataquality/loggers/data_logger/object_detection.py
Ben-Epstein: @bogdan-galileo that is the type for the column name/index. This is the value

dataquality/integrations/ultralytics.py
Ben-Epstein: great catch!!

Ben-Epstein: updated

bogdan-galileo (approved):
LGTM, thanks!!

PR #589, opened 2023-05-11
Added locks in order to allow logging in multithreaded way

dcaustin33:
Added locks to the semantic segmentation data logger as we are logging in a multithreaded way that has the potential to try to access and write to the same file for different dataframes.

### DIFF
--- a/docs/cv/coco_deeplab_hooks.ipynb
+++ b/docs/cv/coco_deeplab_hooks.ipynb
@@ -2,20 +2,9 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "'ds = load_dataset(\\n    \"CVdatasets/CocoSegmentationOnlyVal5000\",\\n    use_auth_token=\"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\"\\n)'"
-      ]
-     },
-     "execution_count": 1,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
     "# !pip install datasets evaluate torch torchvision \n",
     "import os\n",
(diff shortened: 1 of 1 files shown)


Ben-Epstein (commented):
hey Derek, can we chat through this a bit? I feel like we can clean this up and remove some duplication

dataquality/loggers/data_logger/semantic_segmentation.py
```diff
@@ -114,6 +117,43 @@ def log_image_samples(
...
+        lock.acquire()
+        file_path = self.input_data_file()
+        if self.log_export_progress:
+            with vaex.progress.tree("vaex", title=f"Logging {len(df)} samples"):
+                df.export(file_path)
+        else:
+            df.export(file_path)
+
+        df.close()
+        self.logger_config.input_data_logged[str(self.split)] += 1
```
Ben-Epstein, line 154: lets break this part out in the base logger so we can override just this in the child (and call `super` in a lock)

dataquality/loggers/data_logger/semantic_segmentation.py
```diff
@@ -114,6 +117,43 @@ def log_image_samples(
...
+        os.makedirs(f"{self.input_data_path}/{self.split}", exist_ok=True)
+
+        df = self._get_input_df()
+        # Validates cloud size limit
+        self.validate_data_size(df)
+
+        ids = df["id"].tolist()
+        self.validate_ids_for_split(ids)
+        self.add_ids_to_split(ids)
+        lock.acquire()
```
Ben-Epstein, line 145: we should use `with lock` isntead

dataquality/loggers/data_logger/semantic_segmentation.py
```diff
@@ -24,6 +26,7 @@
 # containing image data often won't fit in memory
 
 ITER_CHUNK_SIZE_IMAGES = 1000
+lock = threading.Lock()
```
Ben-Epstein, line 29: we already have a lock in the system. Let's import and use that one? https://github.com/rungalileo/dataquality/blob/main/dataquality/utils/thread_pool.py#L9

dataquality/loggers/data_logger/semantic_segmentation.py
Ben-Epstein: see here https://github.com/rungalileo/dataquality/blob/14bc7cb3661ba84d74b98e89e8ae69def7e19296/dataquality/utils/hdf5_store.py#L75

dataquality/loggers/data_logger/semantic_segmentation.py
```diff
@@ -114,6 +115,24 @@ def log_image_samples(
...
+        with lock:
+            file_path = self.input_data_file()
+            if self.log_export_progress:
+                with vaex.progress.tree("vaex", title=f"Logging {len(df)} samples"):
+                    df.export(file_path)
+            else:
+                df.export(file_path)
+
+            df.close()
+            self.logger_config.input_data_logged[str(self.split)] += 1
```
Ben-Epstein, line 134:
```suggestion
            super().export_df(df)
```

elboy3 (approved):
delicious

PR #590, opened 2023-05-11
Bug/extra inference labels validation

Ben-Epstein:
When someone trains a model, logging data to galileo, we validate that the shape of their labels matches the shape of their logits (for indexing and such).

BUT, if the user comes back the next day and does an inference run over that same run, but with new labels, and new logits, we don't validate that it matches the existing training data logged. This is problematic and will break things in the API. 

This addresses that problem

image

### DIFF
--- a/dataquality/clients/api.py
+++ b/dataquality/clients/api.py
@@ -283,9 +283,7 @@ class ApiClient:
         project, run = self._get_project_run_id(
             project_name=project_name, run_name=run_name
         )
-        task_type = self.get_task_type(project, run)
-        labels = Route.ner_labels if task_type == TaskType.text_ner else Route.labels
-        url = f"{config.api_url}/{Route.content_path(project, run)}/{labels}"
+        url = f"{config.api_url}/{Route.content_path(project, run)}/{Route.labels}"
         params = {"task": task} if task else None
         res = self.make_request(RequestType.GET, url=url, params=params)
         return res["labels"]
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -104,8 +104,8 @@ def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
         dataquality.set_labels_for_run(labels)
         dataquality.get_data_logger().logger_config.existing_run = True
         print(
-            f"🚀 Found existing run labels. Setting labels for run to {labels}. Do not "
-            f"set labels for this run."
+            f"🚀 Found existing run labels. Setting labels for run to {labels}. You do "
+            f"not need to set labels for this run."
         )
     # We except all Exceptions here because this should _never_ prohibit the user from
     # creating a new run.
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -88,6 +88,22 @@ class InitManager:
             if not os.path.exists(out_dir):
                 os.makedirs(out_dir)
 
+def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
+    """When a run already exists, fetch and set the labels automatically
+
+    When users call `dq.init` with a project/run that already exists (for example,
+    to run inference), we want to fetch their labels for them and set them as a
+    convenience.
+
+    We also set the `existing_run` attribute to be True, so they cannot change their
+    labels. Their inference labels must match their train/test/val split labels
+    """
+    pass
+    # try:
+    #     labels = dataquality.metrics.get_labels_for_run(project_name, run_name)
+    #     dataquality.set_labels_for_run(labels)
+    #     dataquality.get_data_logger().logger_config.existing_run = True
+    # except
 
 @check_noop
 def init(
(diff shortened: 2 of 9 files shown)


dataquality/core/init.py
```diff
@@ -89,6 +89,30 @@ def create_log_file_dir(
...
 
+def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
+    """When a run already exists, fetch and set the labels automatically
+
+    When users call `dq.init` with a project/run that already exists (for example,
+    to run inference), we want to fetch their labels for them and set them as a
+    convenience.
+
+    We also set the `existing_run` attribute to be True, so they cannot change their
+    labels. Their inference labels must match their train/test/val split labels
```
setu4993, line 100: Phew!

dataquality/core/init.py
```diff
@@ -89,6 +89,30 @@ def create_log_file_dir(
...
+    """
+    try:
+        labels = dataquality.metrics.get_labels_for_run(project_name, run_name)
+        dataquality.set_labels_for_run(labels)
+        dataquality.get_data_logger().logger_config.existing_run = True
+        print(
+            f"🚀 Found existing run labels. Setting labels for run to {labels}. Do not "
+            f"set labels for this run."
+        )
```
setu4993, line 109:
```suggestion
        print(
            f"🚀 Found existing run labels. Setting labels for run to {labels}. You do not "
            f"need to set labels for this run."
        )
```

dataquality/clients/api.py
```diff
@@ -283,10 +283,6 @@ def get_labels_for_run(
         project, run = self._get_project_run_id(
             project_name=project_name, run_name=run_name
         )
-        task_type = self.get_task_type(project, run)
-        if not task and task_type == TaskType.text_multi_label:
-            raise GalileoException("For multi-label runs, a task name must be provided")
```
setu4993, line 288: Why are we removing this?

dataquality/core/init.py
```diff
@@ -89,6 +89,30 @@ def create_log_file_dir(
...
+    # We except all Exceptions here because this should _never_ prohibit the user from
+    # creating a new run.
+    except Exception:
```
anthonycorletti, line 112: should we print/ warn that an exception occurred and then continue on?

dataquality/core/init.py
```diff
@@ -191,3 +214,12 @@ def init(
...
     dataquality.get_data_logger().logger_config.reset(factory=True)
+    if not run_created:
+        _set_labels_for_existing_run(project_name, run_name)
+
+
+def delete_run(project_name: str, run_name: str) -> None:
+    """Deletes a run from Galileo"""
+    pid, rid = api_client._get_project_run_id(project_name, run_name)
+    api_client.delete_run(pid, rid)
+    print(f"🗑 Run {project_name}/{run_name} deleted")
```
setu4993, line 225: Would this be nicer?

```suggestion
    print(f"🗑 Run {run_name} in project {project_name} deleted.")
```

anthonycorletti (approved):

anthonycorletti (approved):

dataquality/loggers/logger_config/base_logger_config.py
```diff
@@ -34,7 +34,8 @@ class BaseLoggerConfig(BaseModel):
     feature_names: List[str] = []
     metadata_documents: Set = set()  # A document is a large str > 1k chars < 10k chars
     finish: Callable = lambda: None  # Overwritten in Semantic Segmentation
-    polygon_idx: int = 0
+    # True when calling `init` with a run that already exists
+    existing_run: bool = False
```
setu4993, line 38: 💯

dataquality/loggers/logger_config/semantic_segmentation.py
```diff
@@ -2,7 +2,7 @@
 
 
 class SemanticSegmentationLoggerConfig(BaseLoggerConfig):
-    pass
+    polygon_idx: int = 0
```
setu4993, line 5: Unrelated change?

dataquality/core/init.py
franz101, line 112: should we log the exception? for example with dq.analytics

setu4993 (approved):
Amazing. Thanks for the debug the other day and this fix!

dataquality/clients/api.py
Ben-Epstein: because it's actually handled on the API, so this is no longer necessary

dataquality/loggers/logger_config/semantic_segmentation.py
Ben-Epstein: yea, unrelated but small and easy to do

dataquality/clients/api.py
Ben-Epstein: just noticed it while working through the PR

dataquality/core/init.py
Ben-Epstein: nice. THis will break linting, but ill update it locally and push up

dataquality/core/init.py
Ben-Epstein: i dont think so. The most common error will just be that there weren't in fact any labels for this run (it never had data logged), and that is OK

elboy3 (approved):
very nice 👏

PR #592, opened 2023-05-11
Feature/semseg polygon dep

dcaustin33:
Adds polygon dep as an attribute and compute dep for each polygon. 
Also adds a helper function to polygon_data to make it more readable


dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -133,6 +195,13 @@ def _get_data_dict(self) -> Dict:
         calculate_misclassified_polygons_batch(self.pred_masks, gold_polygons_batch)
         calculate_undetected_polygons_batch(self.pred_masks, gold_polygons_batch)
 
+        calculate_dep_polygons_batch(
+            gold_polygons_batch,
+            dep_heatmaps.numpy(),
+            height=[img.shape[-1] for img in self.gold_masks],
+            width=[img.shape[-2] for img in self.gold_masks],
```
elboy3, line 202: this is duplicated on 210 below, lets just call it once and reuse var

dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -157,50 +226,12 @@ def _get_data_dict(self) -> Dict:
...
-        polygon_data = {
-            "id": polygon_ids,
-            "image_id": image_ids,
-            "pred": preds,
-            "gold": golds,
-            "data_error_potential": data_error_potentials,
-            "galileo_error_type": errors,
-            "split": [self.split] * len(image_ids),
-        }
+        polygon_data = self.get_polygon_data(pred_polygons_batch, gold_polygons_batch)
```
elboy3, line 229:
```suggestion
        polygon_data = self.get_polygon_data(pred_polygons_batch, gold_polygons_batch)
        n_polygons = polygon_data["image_id"]
```

i think more clear if we do this

elboy3 (approved):
looks nice! can you create a new run called `polygon-dep` to test before you merge? 

i can take a look at the processed DF and make sure it saved right

PR #594, opened 2023-05-11
truly random names using uuid

Ben-Epstein:
customers pointed out that sometimes the random names were being repeated. This happened because they were setting seeds in their code, which would cause the random generator to reset and return the same name.

UUIDs are not affected by the seed in the same way (tbh not sure how thats possible), but this prevents the issue from occuring. We tac on 5 characters from a uuid to make sure the name is random in the case of a seed being set

### DIFF
--- a/tests/utils/test_name.py
+++ b/tests/utils/test_name.py
@@ -72,3 +72,4 @@ def test_name_random_seed() -> None:
     for _ in range(5):
         random.seed(0)
         names.append(random_name())
+    assert len(set(names)) == len(names)
--- a/tests/utils/test_name.py
+++ b/tests/utils/test_name.py
@@ -66,10 +66,11 @@ def test_name() -> None:
     assert anm in ANIMALS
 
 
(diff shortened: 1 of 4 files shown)


tests/utils/test_name.py
```diff
@@ -62,3 +64,12 @@ def test_name() -> None:
...
     assert anm in ANIMALS
+
+
+def test_name_random_seed() -> None:
+    """Setting a seed should still let us have random names"""
+    names = []
+    for _ in range(5):
+        random.seed(0)
+        names.append(random_name())
+    assert len(set(names)) == len(names)
```
setu4993, line 75: Might be worth testing this with a few different seed values (`mark.parameterize`).

setu4993 (approved):

tests/utils/test_name.py
Ben-Epstein: added

setu4993 (approved):
💯

PR #600, opened 2023-05-15
Fix Iterable Index error

franz101:
- Quickest fix in history of DQ
Turns out the error was caused because the indices were a generator not a list.
Added test

### DIFF
--- a/dataquality/integrations/fastai.py
+++ b/dataquality/integrations/fastai.py
@@ -278,6 +278,11 @@ class FastAiDQCallback(Callback):
         # Store the current batch ids by trimming the stored ids by
         # the batch size length
         cur_split = self.logger_config.cur_split
+        indices = self.idx_store[FAIKey.dataloader_indices][cur_split][-1]
+        # check if list or iterator and convert to list
+        if not isinstance(indices, list):
+            self.idx_store[FAIKey.dataloader_indices][cur_split][-1] = indices
+
         indices = self.idx_store[FAIKey.dataloader_indices][cur_split][-1][
             :bs_len
         ].copy()
--- a/dataquality/integrations/fastai.py
+++ b/dataquality/integrations/fastai.py
@@ -281,7 +281,7 @@ class FastAiDQCallback(Callback):
         indices = self.idx_store[FAIKey.dataloader_indices][cur_split][-1]
         # check if list or iterator and convert to list
         if not isinstance(indices, list):
-            self.idx_store[FAIKey.dataloader_indices][cur_split][-1] = indices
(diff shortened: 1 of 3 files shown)


dataquality/integrations/fastai.py
```diff
@@ -283,6 +278,11 @@ def after_pred(self) -> None:
         # Store the current batch ids by trimming the stored ids by
         # the batch size length
         cur_split = self.logger_config.cur_split
+        indices = self.idx_store[FAIKey.dataloader_indices][cur_split][-1]
+        # check if list or iterator and convert to list
+        if not isinstance(indices, list):
+            self.idx_store[FAIKey.dataloader_indices][cur_split][-1] = indices
```
Ben-Epstein, line 284: shouldnt this be
```suggestion
            self.idx_store[FAIKey.dataloader_indices][cur_split][-1] = list(indices)
```

tests/integrations/fastai/test_text_fai.py
```diff
@@ -0,0 +1,95 @@
...
+    mock_get_project_by_name: MagicMock,
+    set_test_config: Callable,
+    mock_wait_for_run: MagicMock,
+    mock_make_request: MagicMock,
+    mock_upload_log_file: MagicMock,
+    mock_reset_run: MagicMock,
+    mock_version_check: MagicMock,
+    cleanup_after_use: Generator,
+) -> None:
+    global df
```
setu4993, line 66: My 0.02: We should avoid using global dfs for tests.

tests/integrations/fastai/test_text_fai.py
```diff
@@ -0,0 +1,95 @@
...
+    mock_get_project_by_name.return_value = {"id": DEFAULT_PROJECT_ID}
+    mock_create_run.return_value = {"id": DEFAULT_RUN_ID}
+    set_test_config(current_project_id=None, current_run_id=None)
+    dls = TextDataLoaders.from_df(
+        df, text_col="text", label_col="label", drop_last=False, bs=2
+    )
+    dq.init(task_type=TaskType.text_classification)
+    labels = dls.vocab[-1]
+    print("labels")
+    print(labels)
```
setu4993, line 76: I don't think we want prints in tests?
```suggestion
```

setu4993 (approved):
Left 1 note about using a global DF, but rest seems fine to me.

tests/integrations/fastai/test_text_fai.py
franz101: sure thing

PR #601, opened 2023-05-15
nicer naming system for runs

Ben-Epstein:
the core of this pr is the new `create_run_name` function + tests for it. 

User feedback suggested a nicer naming system. Instead of random names that are hard to parse, create run names that are today's date followed by the run_number that was created today.

tested with pytest, as well as:
With a new project
image

And an existing project
image

### DIFF
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -122,11 +122,11 @@ def create_run_name(project_name: str) -> str:
     this project with this scheme.
 
     ie:
-        2023-05-15-1
-        2023-05-15-2
-        2023-05-15-3
+        2023-05-15_1
+        2023-05-15_2
+        2023-05-15_3
         ...
-        2023-05-15-n
+        2023-05-15_n
     """
     pid = api_client.get_project_by_name(project_name)["id"]
     runs = api_client.get_project_runs(pid)
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -136,12 +136,12 @@ def create_run_name(project_name: str) -> str:
         # If it's not an auto-created run name, skip
         if today not in run["name"]:
             continue
-        splitter = f"{today}-"
+        splitter = f"{today}_"
         run_num = run["name"].split(splitter)[-1]
         if run_num.isdigit() and int(run_num) > max_run_today:
             max_run_today = int(run_num)
     run_num = max_run_today + 1
-    return f"{today}-{run_num}"
+    return f"{today}_{run_num}"
 
 
 @check_noop
@@ -194,11 +194,11 @@ def init(
         return
 
     project_name = validate_name(project_name, assign_random=True)
+    project, proj_created = _init.get_or_create_project(project_name, is_public)
+
     if not run_name:
         run_name = create_run_name(project_name)
     run_name = validate_name(run_name, assign_random=False)
-
-    project, proj_created = _init.get_or_create_project(project_name, is_public)
     run, run_created = _init.get_or_create_run(project_name, run_name, task_type)
     if not run_created:
         warnings.warn(
--- a/dataquality/core/init.py
+++ b/dataquality/core/init.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import warnings
+from datetime import datetime
 from typing import Dict, Optional, Tuple
 
 from pydantic.types import UUID4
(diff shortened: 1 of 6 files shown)


anthonycorletti (approved):
:shipit: 

on another note, this makes me think that searching by project name in the search pane might be useful.

franz101:
b-e-a-t-i-f-u-l

dataquality/core/init.py
```diff
@@ -113,6 +114,36 @@ def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
...
+    for run in runs:
+        # If it's not an auto-created run name, skip
+        if today not in run["name"]:
+            continue
+        splitter = f"{today}_"
+        run_num = run["name"].split(splitter)[-1]
+        if run_num.isdigit() and int(run_num) > max_run_today:
+            max_run_today = int(run_num)
+    run_num = max_run_today + 1
+    return f"{today}_{run_num}"
```
setu4993, line 144: Nit: It might be nice to differentiate between user created names and auto-generated names. Maybe a simple prefix like `auto-2023...`.

Given we use dates, I can totally imagine users (like me!) who prefer to organize it by dates themselves.

dataquality/utils/name.py
```diff
@@ -1612,9 +1612,9 @@ def validate_name(name: Optional[str], assign_random: bool = False) -> str:
 
 def random_name() -> str:
     result = [
-        str(uuid4())[:5],
         random.choice(ADJECTIVES),
         random.choice(COLORS),
         random.choice(ANIMALS),
+        str(uuid4())[:5],
```
setu4993, line 1618: Why did we move this to the end?

tests/core/test_init.py
```diff
@@ -36,7 +39,9 @@
 @patch.object(ApiClient, "create_run")
 @patch("dataquality.core.init._check_dq_version")
 @patch.object(dataquality.core.init.ApiClient, "valid_current_user", return_value=True)
+@patch("dataquality.core.init.create_run_name", return_value="foo")
```
setu4993, line 42: Nit: Given we're repeating this a few times in here, would definitely be nicer as a fixture.

dataquality/core/init.py
```diff
@@ -113,6 +114,36 @@ def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
...
+    If a run_name is not passed into `init`, we create a run_name base with today's
+    date, and increment the digit at the end based on how many runs were created in
+    this project with this scheme.
+
+    ie:
+        2023-05-15-1
+        2023-05-15-2
+        2023-05-15-3
+        ...
+        2023-05-15-n
```
setu4993, line 129: Looks like we are using `-` as the delimiter. In which case, the docstring is incorrect...
```suggestion
    ie:
        2023-05-15_1
        2023-05-15_2
        2023-05-15_3
        ...
        2023-05-15_n
```

setu4993 (approved):
Looks good. Left a few notes.

dataquality/core/init.py
```diff
@@ -113,6 +114,36 @@ def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
...
+    """
```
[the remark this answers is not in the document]
Ben-Epstein: Yea random had a few issues, and the only true feedback from customers has been they prefer this

dataquality/core/init.py
```diff
@@ -113,6 +114,36 @@ def _set_labels_for_existing_run(project_name: str, run_name: str) -> None:
...
+    pid = api_client.get_project_by_name(project_name)["id"]
+    runs = api_client.get_project_runs(pid)
+    today = datetime.today().strftime("%Y-%m-%d")
+    max_run_today = 0
+    for run in runs:
+        # If it's not an auto-created run name, skip
+        if today not in run["name"]:
+            continue
+        splitter = f"{today}_"
```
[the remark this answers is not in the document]
setu4993: How do you feel about `auto-YYYYMMDD-i`?

dataquality/utils/name.py
Ben-Epstein: Just arbitrary, I thought the uuid in the beginning looked pretty bad. And since we made that change just a few days ago, I don't think anyone's experienced it yet

dataquality/core/init.py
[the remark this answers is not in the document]
Ben-Epstein: We can add auto, but it'll increase the complexity of the parsing. I'd be surprised if users date things exactly the same way as us?

tests/core/test_init.py
Ben-Epstein: Yea good point

dataquality/core/init.py
Ben-Epstein: Nice catch. I changed it halfway through, missed the docstring

dataquality/core/init.py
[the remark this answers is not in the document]
setu4993: FWIW, `YYYY-MM-DD` is a pretty standard structure and the default for `str(date.today())`.

PR #606, opened 2023-05-17
Polygon uuid as strs 

elboy3:
whoops

### DIFF
--- a/dataquality/schemas/semantic_segmentation.py
+++ b/dataquality/schemas/semantic_segmentation.py
@@ -47,7 +47,7 @@ class Contour(BaseModel):
 
 
 class Polygon(BaseModel):
-    uuid: UUID4
+    uuid: str  # UUID4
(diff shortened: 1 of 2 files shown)


dcaustin33 (approved):

dataquality/schemas/semantic_segmentation.py
```diff
@@ -47,7 +47,7 @@ class Contour(BaseModel):
 
 
 class Polygon(BaseModel):
-    uuid: UUID4
+    uuid: str  # UUID4
```
setu4993, line 50: This with `UUID4` should work since we're passing in the `str` version of it from `polygons.py`, right?

setu4993 (commented):
Left a question... Is it easy to add a test for this?

anthonycorletti (approved):

Ben-Epstein (approved):

PR #607, opened 2023-05-17
Feat/ic support image folder

bogdan-galileo:
>
>

***Demonstration***


dataquality/loggers/data_logger/image_classification.py
```diff
@@ -72,6 +75,32 @@ def log_image_dataset(
         column_map: Optional[Dict[str, str]] = None,
         parallel: bool = False,
     ) -> Any:
+        """
+        Log an image dataset of input samples for image classification
```
Ben-Epstein, line 79: @bogdan-galileo this docstring is great, but Can you move this up to the top level in `core/log.py::log_image_dataset`[here](https://github.com/rungalileo/dataquality/pull/607/files#diff-c359f7d9115e658179c9113608b1910385ce58010df9c04cc0cf07a1c617eadeR134) -- this is perfect but should be at the top level (so the user can call `help(dq.log_image_dataset)`

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -194,6 +223,28 @@ def _prepare_hf(
 
         return prepared
 
+    def _prepare_df_from_ImageFolder(
```
Ben-Epstein, line 226: can we add a test for this function please

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: Cool thanks, I wasn't sure where we wanted it, makes sense.

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: done!

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -194,6 +204,28 @@ def _prepare_hf(
...
+        self,
+        dataset: "ImageFolder",  # type: ignore # noqa: F821
+        imgs_remote_location: Optional[str] = None,
+    ) -> pd.DataFrame:
+        """
+        Create a dataframe containing the ids, labels and paths of the images
+        coming from an ImageFolder dataset.
+        """
+        df = pd.DataFrame(columns=["text", "label"], data=dataset.imgs)
```
franz101, line 216: what is the difference between samples and images?

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -194,6 +204,28 @@ def _prepare_hf(
...
+        label_idx_to_label = {
+            label_idx: label for label, label_idx in dataset.class_to_idx.items()
+        }
+        df["label"] = df.label.map(label_idx_to_label)
+        df = df.reset_index().rename(columns={"index": "id"})
+
+        # Replace the paths with the remote one, if a remote location is specified
+        if imgs_remote_location is not None:
+            df["text"] = df["text"].str.replace(dataset.root, imgs_remote_location)
```
franz101, line 225: does this work for
root= ./path
root= /path
imgs_remote_location = ./path
imgs_remote_location = /path

just asking because in yolo the relative path was causing problems and had to be resolved with path lib or os.cwd

tests/assets/train_images/labelB/image4.png
franz101, line 1: love this test data :D

franz101 (approved):
Looks good to me, just a couple of questions around .samples vs .images

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: Did I use both ? They're the same, see last line in https://pytorch.org/vision/main/_modules/torchvision/datasets/folder.html#ImageFolder

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: Good point. Just looked and seems to not be a problem since
- the path in dataset.imgs always starts with dataset.root (i.e., is not resolved)
- imgs_remote_location will be a GC or S3 uri of the type gs:// or s3://

So the swap always replaces the beginning of the local path (relative or absolute) and we always get an absolute remote path. Did you have something else in mind ? @franz101

PR #608, opened 2023-05-17
new reprocess run

Ben-Epstein:
the old one was pretty outdated, and was missing the ability to get `ner_labels`, which made it break NER runs.

This one fixes that, and has a new home (because we needed to access `dq.metrics`

working ner run
image

good failure case (which tells you what to run) for the old function
image

### DIFF
--- a/dataquality/clients/api.py
+++ b/dataquality/clients/api.py
@@ -334,70 +334,12 @@ class ApiClient:
         labels: Optional[Union[List, List[List]]] = None,
         xray: bool = False,
     ) -> Dict:
-        """Reinitiate a project/run that has already been finished
-
-        If a project and run name have been provided, that project/run will be
-        reinitiated, otherwise we trigger the currently initialized project/run.
-
-        This will clear out the current state in the server, and will recalculate
-        * DEP score
-        * UMAP Embeddings for visualization
-        * Smart features
-
-        :param project_name: If not set, will use the currently active project
-        :param run_name: If not set, will use the currently active run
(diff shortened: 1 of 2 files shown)


dataquality/clients/api.py
```diff
@@ -334,70 +334,12 @@ def reprocess_run(
...
-            task_type=task_type,
-            xray=xray,
-        )
-        res = self.make_request(
-            RequestType.POST, url=f"{config.api_url}/{Route.jobs}", body=body
+        """Removed. Please see dq.internal.reprocess_run"""
+        raise GalileoException(
+            "It seems you are trying to reprocess a run. Please call the following:\n\n"
+            "from dataquality.internal import reprocess_run\n\n"
+            f"reprocess_run('{project_name}', '{run_name}')"
```
setu4993, line 341: Nice clean-up.

dataquality/internal.py
```diff
@@ -0,0 +1,67 @@
...
+                f"It seems no data is available for run " f"{project_name}/{run_name}"
+            )
+        raise e from None
+    # There were no labels available for this run
+    except KeyError:
+        raise GalileoException(
+            "It seems we cannot find the labels for this run. This means "
+            "that the run cannot be reprocessed, because it likely was never processed "
+            "to begin with. Please call dq.init() and re-train your model "
+        ) from None
```
setu4993, line 43: What's the `from None` at the end here 🤔?

setu4993 (approved):
Thanks for the quick fix!

PR #603, opened 2023-05-16
Improve YOLO test speed

franz101:
Instead of training on coco128 we create a custom dataset and we only have 4 samples

PR #613, opened 2023-05-18
Ultralytics: fix rel path issue during predict

bogdan-galileo:
For ultralytics integration (Object Detection). 

When calling predict with relative paths in the data.YAML file, our `dqyolo` integration creates a temporary yaml file stored at a different location. The images rel paths don't make sense anymore, see below for example where yaml is saved in `/tmp`

Screenshot 2023-05-18 at 2 21 54 PM

We overwrite the rel paths with absolute paths. 
Tested all scenarios where yaml file is given as abs/rel path, path in yaml file is given as abs/rel path.


bogdan-galileo:
Closing, fix added to https://github.com/rungalileo/dataquality/pull/603

PR #614, opened 2023-05-18
Tiny change to make misclassified compute correctly

dcaustin33:
Tiny change to make sure misclassified is not including background predictions


.gitignore
```diff
@@ -178,7 +178,7 @@ coco_dataset/
 coco.ipynb
 
 # SemSeg
+coco_deeplab_hooks.ipynb
```
setu4993, line 181: We should just ignore all notebooks.

setu4993 (approved):

.gitignore
setu4993: Oh never mind 🤦🏽. Just noticed we do have notebooks [in here](https://github.com/rungalileo/dataquality/tree/main/docs/notebooks).

PR #622, opened 2023-05-22
Small/make list

Ben-Epstein:


setu4993 (approved):

dataquality/core/log.py
```diff
@@ -485,7 +485,7 @@ def set_labels_for_run(labels: Union[List[List[str]], List[str]]) -> None:
     task-order of the task-probabilities logged as well.
     """
     if get_data_logger().logger_config.existing_run:
-        if sorted(labels) != sorted(get_data_logger().logger_config.labels):
+        if sorted(list(labels)) != sorted(get_data_logger().logger_config.labels):
```
setu4993, line 488: If we know this is going to be a `pd.Series`, we should probably use:

```suggestion
        if sorted(labels.to_list()) != sorted(get_data_logger().logger_config.labels):
```

Ben-Epstein: we don't know. It should be a list, but the user can technically pass anything (numpy array, series, set, etc).

I found this issue because I accidentally passed in a pandas series

setu4993: Yea, just worried that it's not 100% so we might see this break. If you feel confident about this change, go for it.

PR #627, opened 2023-05-22
Feature: include LM percentage per polygon (#4)

dcaustin33:
Calculate LM percentage per ground truth polygon


elboy3 (approved):
I'm labelllling this as an excellllent PR

PR #628, opened 2023-05-22
Feature/semseg misclassification

dcaustin33:
Adds misclassification to semseg polygons

### DIFF
--- a/dataquality/loggers/model_logger/semantic_segmentation.py
+++ b/dataquality/loggers/model_logger/semantic_segmentation.py
@@ -134,6 +134,9 @@ class SemanticSegmentationModelLogger(BaseGalileoModelLogger):
         background_error_pcts = []
         polygon_areas = []
         lm_pcts = []
+        polygon_accuracy = []
+        mislabled_class = []
+        mislabled_class_pct = []
         for i, image_id in enumerate(self.image_ids):
             pred_polygons = pred_polygons_batch[i]
             for polygon in pred_polygons:
@@ -145,6 +148,13 @@ class SemanticSegmentationModelLogger(BaseGalileoModelLogger):
                 background_error_pcts.append(polygon.background_error_pct)
                 polygon_areas.append(polygon.area)
                 lm_pcts.append(polygon.likely_mislabeled_pct)
+                polygon_accuracy.append(polygon.classification_data.accuracy)
+                mislabled_class.append(
+                    polygon.classification_data.dominant_mislabel_class
+                )
+                mislabled_class_pct.append(
+                    polygon.classification_data.dominant_mislabel_class_percent
+                )
                 upload_polygon_contours(polygon, self.contours_path)
                 polygon_ids.append(polygon.uuid)
             gold_polygons = gold_polygons_batch[i]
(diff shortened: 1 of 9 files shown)


dataquality/schemas/semantic_segmentation.py
```diff
@@ -15,6 +15,20 @@ class SemSegCols(str, Enum):
...
+    accuracy: float the mean accuracy per pixel
+    dominant_mislabel_class: int the non ground truth class that had the most pixels
+       in the polygon
+    dominant_mislabel_class_percent: float the percentage of pixels in the polygon
+        that were classified as mislabel_class
+    """
+
+    accuracy: Optional[float] = None
+    dominant_mislabel_class: Optional[int] = None
+    dominant_mislabel_class_percent: Optional[float] = None
```
elboy3, line 29: if these are all present when we instantiate the object, we don't need them to be optional

dataquality/schemas/semantic_segmentation.py
```diff
@@ -15,6 +15,20 @@ class SemSegCols(str, Enum):
     meta = "meta"
 
 
+class ClassificationData(BaseModel):
```
elboy3, line 18: Maybe `ClassificationErrorData`

dataquality/schemas/semantic_segmentation.py
```diff
@@ -15,6 +15,20 @@ class SemSegCols(str, Enum):
...
+    """
+    accuracy: float the mean accuracy per pixel
```
elboy3, line 20: what does "mean accuracy per pixel" mean? specifically the word "mean" in this context

dataquality/schemas/semantic_segmentation.py
```diff
@@ -30,7 +44,7 @@ class IoUType(str, Enum):
 class IouData(BaseModel):
     iou: float
     iou_per_class: List[float]
-    area_per_class: List[int]
+    area_per_class: List[float]
```
elboy3, line 47: if this is a float are the other area values a float? if so we should update the fn typing to return float not int

dataquality/schemas/semantic_segmentation.py
```diff
@@ -62,7 +76,7 @@ class Contour(BaseModel):
 class Polygon(BaseModel):
     uuid: str  # UUID4
     label_idx: int
-    misclassified_class_label: Optional[int] = None
+    classification_data: ClassificationData = ClassificationData()
```
elboy3, line 79: just have this be optional and default to None

dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -208,7 +228,12 @@ def _get_data_dict(self) -> Dict:
             self.pred_masks, self.gold_masks
         )
         # Errors
-        add_classification_error_to_polygons_batch(self.pred_masks, gold_polygons_batch)
+        add_classification_error_to_polygons_batch(
+            self.pred_masks, gold_polygons_batch, n_classes
+        )
+        add_classification_error_to_polygons_batch(
+            self.gold_masks, pred_polygons_batch, n_classes
+        )
```
elboy3, line 236: why are we doing it for both now?

elboy3:
To fix typing issue in 3.8 and 3.9 https://stackoverflow.com/questions/76313592/import-langchain-error-typeerror-issubclass-arg-1-must-be-a-class

PR #634, opened 2023-05-26
Feature/contour speed optimizations

dcaustin33:
Optimizations to write contours all locally and then upload in one file at the end because of colab's speed. For 50 examples this takes us from 60 seconds down to 30 seconds in colab.

### DIFF
--- a/dataquality/utils/semantic_segmentation/polygons.py
+++ b/dataquality/utils/semantic_segmentation/polygons.py
@@ -11,7 +11,6 @@ import torch
 from dataquality.clients.objectstore import ObjectStore
 from dataquality.schemas.ml import ClassType
 from dataquality.schemas.semantic_segmentation import Contour, Pixel, Polygon
-from dataquality.utils.thread_pool import lock
 
 object_store = ObjectStore()
 
--- a/dataquality/utils/semantic_segmentation/polygons.py
+++ b/dataquality/utils/semantic_segmentation/polygons.py
@@ -157,8 +157,8 @@ def write_polygon_contours_to_disk(
         prefix(str): prefix of the object name in storage
             \"{proj_run_path}/{split_name_path}/contours"
     """
-    with lock:
-        os.makedirs(prefix, exist_ok=True)
+
+    os.makedirs(prefix, exist_ok=True)
     local_path = f"{prefix}/{polygon.uuid}.json"
 
     with open(local_path, "w") as f:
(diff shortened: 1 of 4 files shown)


Ben-Epstein:
@dcaustin33 @elboy3 how will retrieval work on the backend with a single file? Are you going to need to download it every time?

Or in runners, will you split this up into individual files?

elboy3:
> @dcaustin33 @elboy3 how will retrieval work on the backend with a single file? Are you going to need to download it every time?
> 
> Or in runners, will you split this up into individual files?

@Ben-Epstein my thought was to keep it as one. We only need to retrieve it for the /rows endpoint (nothing else). I was considering adding it to the cache as well if the file is large and it takes some time to download. wdyt?

Ben-Epstein:
is it currently a single file, or would this change it into a single file from multiple individual files? I would imagine that a single contours file would be pretty large right?

dcaustin33:
> is it currently a single file, or would this change it into a single file from multiple individual files? I would imagine that a single contours file would be pretty large right?

@Ben-Epstein for 200 examples the size of the file rn is 1.4 mb

Ben-Epstein (approved):
makes sense on the dq side, i would watch the performance on the API and consider breaking its into individual polygon JSONs (or testing that behavior alongside) if the performance suffers

dataquality/utils/semantic_segmentation/polygons.py
```diff
@@ -146,26 +146,22 @@ def draw_polygon(polygon: Polygon, shape: Tuple[int, ...]) -> np.ndarray:
...
     Args:
         polygon(Polygon): A Polygon object
         prefix(str): prefix of the object name in storage
-            - /proj-id/run-id/training/contours/1.json
+            \"{proj_run_path}/{split_name_path}/contours"
     """
-    obj_name = f"{prefix}/{polygon.uuid}.json"
+    if not os.path.isdir(prefix):
+        with lock:
+            os.makedirs(prefix, exist_ok=True)
```
elboy3, line 162: can we just do `os.makedirs(prefix, exist_ok=True)` 


or preferably can we just create this folder at the beginning of a run and then not have to do any checks here?

dataquality/utils/semantic_segmentation/polygons.py
```diff
@@ -157,9 +157,8 @@ def write_polygon_contours_to_disk(
         prefix(str): prefix of the object name in storage
             \"{proj_run_path}/{split_name_path}/contours"
     """
-    if not os.path.isdir(prefix):
-        with lock:
-            os.makedirs(prefix, exist_ok=True)
+    with lock:
+        os.makedirs(prefix, exist_ok=True)
```
elboy3, line 161: why do we need the lock?

PR #637, opened 2023-05-31
Remove LM upload

dcaustin33:
We are no longer uploading the LM map to minio so we have removed the upload from the code. I am still keeping the helper functions in case we should decide to backtrack

### DIFF
--- a/dataquality/loggers/model_logger/semantic_segmentation.py
+++ b/dataquality/loggers/model_logger/semantic_segmentation.py
@@ -195,9 +195,6 @@ class SemanticSegmentationModelLogger(BaseGalileoModelLogger):
         """Returns a dictionary of data to be logged as a DataFrame"""
         # DEP & likely mislabeled
         mean_mislabeled = torch.mean(self.mislabled_pixels, dim=(1, 2)).numpy()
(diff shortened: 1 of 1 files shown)


Ben-Epstein (approved):
this is fine but @dcaustin33 it's worth noting that the function is in our commit history, so we can delete it here and always get it back later if we want it (instead of having dead code)

dataquality/utils/semantic_segmentation/lm.py
```diff
@@ -334,45 +334,6 @@ def _calculate_confidence_joint(
...
-    """
-    for i, image_id in enumerate(image_ids):
-        img = (mislabeled_pixels[i].numpy() * 255).astype(np.uint8)
-        img = Image.fromarray(img, mode="L")
-        img = img.resize((64, 64))
-        upload_im(img, prefix, image_id)
-
-
-def upload_im(img: Image.Image, prefix: str, id: int) -> None:
-    """Uploads one self confident image to minio
```
elboy3, line 356: no insecure images in minio ❌

elboy3 (approved):

PR #640, opened 2023-06-01
Changed file structure for more intutive path in docs

dcaustin33:
Currently writing docs for the various watch functions and all watch functions being with integrations but then semantic segmentation being in a nested directory did not make sense. I think we should write each watch as torch_watch, transformers_watch, semseg_torch_watch etc in the init file for ease of use but just doing this for now.


elboy3 (approved):
tests?

PR #646, opened 2023-06-05
hosted runners

Ben-Epstein:


setu4993 (approved):
🚀

scripts/test.sh
```diff
@@ -4,4 +4,4 @@
 
 ./scripts/lint.sh
 
-pytest --cov=dataquality --cov=tests --cov-report=term-missing --cov-report=xml --durations=10 -o console_output_style=progress --disable-warnings ${@}
\ No newline at end of file
+pytest --cov=dataquality --cov=tests --cov-report=term-missing -n auto --cov-report=xml --durations=10 -o console_output_style=progress --disable-warnings ${@}
```
setu4993, line 7: I tried this a couple weeks ago and it doesn't work as-is. It'll likely require a few updates to `conftest.py`.

Ben-Epstein: okay, ill let these fail and see whats wrong! Unless you have time to look @setu4993

anthonycorletti (approved):
ill_allow_it

tests/integrations/spacy/test_spacy_ner.py
```diff
@@ -431,12 +455,14 @@ def test_spacy_inference_only(
...
-
     # Check if all elements in array1 have a close value in array2
-    assert np.all(any_close)
+    if os.getenv("PYTEST_XDIST_WORKER_COUNT"):
+        np.allclose(embs, embs_expected, atol=1e-3)
+    else:
+        is_close = np.isclose(embs[:, np.newaxis], embs_expected, atol=1e-3)
+        # For each element in array1, check if any element in array2 is close
+        any_close = np.any(is_close, axis=1)
+        assert np.all(any_close)
```
Ben-Epstein, line 465: why is it a different comparison based on multiprocessing?

anthonycorletti: When using xdist, the NumPy arrays are serialized and sent to different worker processes for parallel execution. This serialization process creates copies of the arrays for each worker process, and the memory layout may change.

As a result, if you have tests that rely on the exact memory layout of the NumPy arrays, np.all may fail unexpectedly because the memory addresses of the copied arrays are different from the original array. The element-wise comparison may yield different results due to differences in the memory layout.

A way around this is to use other NumPy functions like np.array_equal or np.allclose that perform element-wise comparisons without considering the memory layout.

Ben-Epstein: Wow interesting. If that's the case, I almost think that we can remove the checks within `else` and just always use allclose. 

Thats typically what we do for most other tests, and it would remove this potentially confusing code for others. 

Wdyt?

anthonycorletti: hm i just realized that i forgot to assert that statement. to be continued.

Ben-Epstein: ‼️  wow good catch!!

PR #650, opened 2023-06-06
Fix SetFit notebook issues v1

franz101:
Identified issues:
Issues when running the dq_evaluate due to the automatic removal of the patches

Further investigation:
Essentiallty the cleanup manager removed the patch. The other issue was that it was checking for the config before it was initialized.

### DIFF
--- a/dataquality/integrations/setfit.py
+++ b/dataquality/integrations/setfit.py
@@ -6,6 +6,7 @@ import torch
 import dataquality as dq
 from dataquality.analytics import Analytics
 from dataquality.clients.api import ApiClient
+from dataquality.core.log import get_data_logger
 from dataquality.schemas.split import Split
 from dataquality.utils.patcher import Cleanup, Patch, PatchManager, RefManager
 from dataquality.utils.setfit import log_preds_setfit
@@ -273,6 +274,9 @@ def unwatch(setfit_obj: Optional[Union["SetFitModel", "SetFitTrainer"]]) -> None
     a.log_function("setfit/unwatch")
     setfitmanager = PatchManager()
     setfitmanager.unpatch()
+    helper_data = dq.get_data_logger().logger_config.helper_data
+    if helper_data:
+        helper_data.clear()
 
 
 def watch(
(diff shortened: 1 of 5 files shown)


dataquality/integrations/setfit.py
```diff
@@ -341,11 +346,9 @@ def evaluate(
     dq_hook = SetFitModelHook(model)
     dq_store = dq_hook.store
 
-    helper_data = dq.get_data_logger().logger_config.helper_data
-
     # Unpatch SetFit model after logging (when finished is called)
-    cleanup_manager = RefManager(dq_hook.unpatch)
-    helper_data["cleaner"] = Cleanup(cleanup_manager)
+    # cleanup_manager = RefManager(dq_hook.unpatch)
+    # helper_data["cleaner"] = Cleanup(cleanup_manager)
```
Ben-Epstein, line 351: lets delete if they arent needed

dataquality/utils/setfit.py
```diff
@@ -62,6 +63,8 @@ def log_preds_setfit(
 
     logger_config = dq.get_data_logger().logger_config
     labels = logger_config.labels
+    epoch = 0 if epoch is None else epoch
```
Ben-Epstein, line 66: slightly cleaner. If epoch is None or 0, itll be set to 0, otherwise, itll be kept
```suggestion
    epoch = epoch or 0
```

dataquality/integrations/setfit.py
```diff
@@ -372,7 +376,11 @@ def dq_evaluate(
 
         if column_mapping is not None:
             dataset = _apply_column_mapping(dataset, column_mapping)
-
+        if "id" not in dataset.features:
+            dataset = dataset.map(lambda x, idx: {"id": idx}, with_indices=True)
+        cur_epoch = get_data_logger().logger_config.cur_epoch
+        if epoch is not None:
+            cur_epoch = epoch
```
Ben-Epstein, line 383: should this be persisted? if so,
```suggestion
        if epoch is not None:
            dq.set_epoch(epoch)
        cur_epoch = get_data_logger().logger_config.cur_epoch
```

dataquality/integrations/setfit.py
franz101: sure

dataquality/utils/setfit.py
franz101: genius

dataquality/integrations/setfit.py
franz101: I thinks that's better

Ben-Epstein (approved):

PR #652, opened 2023-06-07
Feature/dep local_speedup

dcaustin33:
Compute dep locally and moves to cuda for 3 second speed up as well as stores locally before uploading at the end.


.gitignore
```diff
@@ -182,5 +182,6 @@ coco_deeplab_hooks.ipynb
 CV_datasets/
 coco_hf_dataset.py
 large_run.ipynb
+*.png
```
elboy3, line 185: can you double check that this won't break any existing tests? 

tests/assets/train_images uses some pngs. maybe ask franz

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -430,6 +429,39 @@ def upload_contours_split(self, split: str) -> None:
             bucket_name=GALILEO_DEFAULT_RESULT_BUCKET_NAME,
         )
 
+    def upload_dep_split(self, split: str) -> None:
+        """Uploads all dep files for a given split to minio
+
+        Args:
+            split (str): split name
+        """
+        from dataquality.utils.upload import chunk_load_then_upload_df
```
elboy3, line 438: why import here instead of top of file?

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -430,6 +429,39 @@ def upload_contours_split(self, split: str) -> None:
...
+
+        model_logger = dq.get_model_logger()
+        project_path = f"{model_logger.LOG_FILE_DIR}/{config.current_project_id}"
+        local_dep_path = f"{project_path}/{config.current_run_id}/{split}/dep"
+
+        files = os.listdir(local_dep_path)
+        for i, file in enumerate(files):
+            file = f"{local_dep_path}/{file}"
+            files[i] = file
```
elboy3, line 447:
```suggestion
        dep_folder_local = f"{project_path}/{config.current_run_id}/{split}/dep"

        dep_paths = []
        for file in os.listdir(local_dep_path):
            dep_paths.append(f"{local_dep_path}/{file}")
```

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -430,6 +429,39 @@ def upload_contours_split(self, split: str) -> None:
...
+        project_id = config.current_project_id
+        run_id = config.current_run_id
+        split = f"{split}"
```
elboy3, line 450: let's move these up to the top of this fn

also looks like split is already a string from the fn definition?

dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -104,9 +104,8 @@ def lm_path(self) -> str:
         return f"{self.proj_run}/{self.split_name_path}/LM"
```
elboy3, line 104: we can get rid of lm_path, right ?

dataquality/utils/semantic_segmentation/lm.py
```diff
@@ -127,6 +127,7 @@ def calculate_mislabled_by_noise(
         class_samples_margin = probs[:, actual_gold_idx] - probs[:, given_gold_idx]
         # We want the top `num_mislabeled` samples with the largest margin
         # for the actual/expected class relative to the given class
+        num_mislabeled = min(num_mislabeled, class_samples_margin.shape[0])
```
elboy3, line 130: do we need this?

dataquality/utils/semantic_segmentation/metrics.py
```diff
@@ -71,29 +72,46 @@ def calculate_dep_heatmaps(
...
     dep_masks = dep_masks.view(mask_size)
-    return dep_masks
+    gold_masks = gold_masks.cpu()
+    return dep_masks.cpu()
+
+
+def write_dep_to_disk(
+    dep_heatmaps: torch.Tensor,
+    image_ids: List[int],
+    prefix: str,
```
elboy3, line 105: more helpful name than prefix pls

dataquality/utils/semantic_segmentation/metrics.py
```diff
@@ -71,29 +72,46 @@ def calculate_dep_heatmaps(
...
+) -> None:
```
elboy3, line 106: docstring

dataquality/utils/upload.py
```diff
@@ -44,11 +50,18 @@ def __init__(
...
+        self.run_id = run_id
+        self.split = split
+        self.folder = folder
 
     def _upload_file_for_project(
         self,
         file_path: str,
         project_id: Optional[UUID4] = None,
+        run_id: Optional[UUID4] = None,
+        split: Optional[str] = None,
```
elboy3, line 64: why are these needed?

dataquality/utils/upload.py
```diff
@@ -60,8 +73,27 @@ def _upload_file_for_project(
...
+        file_suffix = ""
+        if project_id:
+            file_suffix += f"/{str(project_id)}"
+        if run_id:
+            file_suffix += f"/{str(run_id)}"
+        if split:
+            file_suffix += f"/{split}"
+        if folder:
+            file_suffix += f"/{folder}"
+        return file_suffix + "/"
```
elboy3, line 95: let's just use a `folder_suffix` or something and do `{project_id}/{folder_suffix}` instead of building it here

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -430,6 +429,39 @@ def upload_contours_split(self, split: str) -> None:
...
+        chunk_load_then_upload_df(
+            file_list=files,
+            project_id=project_id,
+            run_id=run_id,
+            split=split,
+            folder="dep",
```
elboy3, line 456: i think we should put these all together here as a `folder_suffix` or something like that instead of passing each individually

dataquality/integrations/torch_semantic_segmentation.py
dcaustin33: Lol whoops

dataquality/utils/semantic_segmentation/lm.py
dcaustin33: Yeah otherwise the percentage can be more than what we actually have in the queue

elboy3 (approved):

PR #653, opened 2023-06-08
Make sure classlabel is added if not existing in hf datasets

franz101:
This a fix to gracefully handle missing features.names when values are int


Ben-Epstein (approved):

dataquality/loggers/data_logger/text_classification.py
```diff
@@ -296,6 +298,14 @@ def parse_label(labels: Union[List[int], List[str]]) -> List[str]:
             f"id ({id}) field must be present in dataset"
         )
 
+        # Make sure the class labels are set
+        if (
+            isinstance(dataset[0].get(label), int)
+            and self.logger_config.labels
+            and not isinstance(dataset.features[label], ClassLabel)
```
Ben-Epstein, line 305: why only if an int?
```suggestion
            self.logger_config.labels
            and not isinstance(dataset.features[label], ClassLabel)
```

Ben-Epstein (approved):

PR #655, opened 2023-06-08
chore: faster build install

anthonycorletti:
* [x] I have updated `docs/`, if necessary.

No issue 🎩 👉 🐇 

locally and ci

- build system: setuptools for consistency with other projects, simplicity, and stability
- consolidated all? configs to pyproject.toml
- replaced bash scripts with invoke to be more pythonic 🐍


CONTRIBUTING.md
```diff
@@ -28,33 +28,26 @@ $ which python pip
 
 ## Flit
 
-This project uses `flit` to manage our project's dependencies.
+This project uses `setuptools` to manage our project's dependencies and `invoke` for running tasks.
```
Ben-Epstein, line 31: see ^^ line 29 still says `Flit`

Ben-Epstein (approved):

dataquality/py.typed
franz101, line 1: what is this file for?

anthonycorletti: https://peps.python.org/pep-0561/

pyproject.toml
```diff
@@ -1,16 +1,18 @@
...
-author = "Galileo Technologies, Inc"
-author-email = "team@rungalileo.io"
-home-page = "https://www.github.com/rungalileo/dataquality"
-description-file = "README.md"
-requires-python = ">=3.7"
 requires = [
+    "setuptools>=65.0",
+    "wheel>=0.37",
+]
+build-backend = "setuptools.build_meta"
```
setu4993, line 7: If we are modifying the build system, we should go the route of using something that's going to provide better checks for our dependency structure (like `poetry`) instead of `setuptools`.

setu4993 (commented):
If we are open to updating the build system, I do strongly think we should go the route of safer build systems like `poetry` which provide us the type of checks we need for packages like `dataquality`. A few features that would help:
- Strict dependency resolution.
- Easy management of dependencies across extras.
- Simple dependency upgrades.

pyproject.toml
anthonycorletti: good thought! the point of this wasnt the tool but to enable some faster build and install workflow in ci. open to improvements.

PR #673, opened 2023-06-14
Fix for SetFit Auto

franz101:
SetFit auto was only working for train and eval, now added training arguments and support for inference and test

### DIFF
--- a/dataquality/integrations/setfit.py
+++ b/dataquality/integrations/setfit.py
@@ -80,7 +80,8 @@ def watch(
 
     pm = PatchManager()
     pm.unpatch()
-    if not previously_initialized:
+
+    if not previously_initialized or (not dq.config.task_type and not project_name):
         init_kwargs: Dict[str, Any] = {}
         if project_name:
             init_kwargs["project_name"] = project_name
--- a/dataquality/integrations/setfit.py
+++ b/dataquality/integrations/setfit.py
@@ -10,6 +10,7 @@ import dataquality as dq
 from dataquality.analytics import Analytics
 from dataquality.clients.api import ApiClient
 from dataquality.core.log import get_data_logger
+from dataquality.dq_auto.auto import AUTO_PROJECT_NAME
 from dataquality.dq_auto.text_classification import (
     TCDatasetManager,
     _get_labels,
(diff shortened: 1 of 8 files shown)


setu4993 (commented):
Nice fix, looks good.

Can we add / update some tests here before we merge?

dataquality/integrations/setfit.py
```diff
@@ -179,84 +182,87 @@ def auto(
...
+    Given a pandas dataframe, a file path, or a Huggingface dataset path, this
+    function will load the data, train a Huggingface transformer model, and
+    provide insights via a link to the Console.
+
+    At least one of `hf_data`, `train_data` should be provided. If neither of
+    those are, a demo dataset will be used for training.
+
+    Parameters
+    ----------
+    hf_data : Union[DatasetDict, str], optional
```
Ben-Epstein, line 205: does this make sense for setfit? Shouldn't the model have to be a SetFitModel or SetFitTrainer?

dataquality/integrations/setfit.py
```diff
@@ -308,19 +314,60 @@ def auto(
...
+            TaskType.text_classification, project_name=project_name, run_name=run_name
+        )
     dq.set_labels_for_run(labels)
-    _log_dataset_dict(dd)
-    trainer, encoded_data = get_trainer(
-        dd, labels, hf_model, max_padding_length, num_train_epochs
+    if not isinstance(hf_model, str):
+        return do_model_eval(hf_model, dd, wait, create_data_embs)
+
+    trainer, encoded_data = get_trainer(dd, hf_model, training_args)
```
Ben-Epstein, line 333: given this line, how does the line above `hf_model: str = "sentence-transformers/paraphrase-mpnet-base-v2",`
make sense as the default? SHouldn't it have to be a SetFitTrainer or SetFitModel?

franz101:
> Nice fix, looks good.
> 
> Can we add / update some tests here before we merge?

Added a test for DQ auto

Ben-Epstein (approved):

dataquality/integrations/setfit.py
franz101: Changed naming 👍

dataquality/integrations/setfit.py
franz101: If a string is provided we finetune the model

PR #674, opened 2023-06-15
enforce order and fail gracefully

Ben-Epstein:


elboy3 (approved):
looks good. why `cur_run_labels` ? for the customer? maybe something less abbreviaty if cust facing `get_run_labels` or `get_current_run_labels`

PR #675, opened 2023-06-15
Remove AG_NEWS dependency

franz101:
Create Mockdataset instead and remove AG_NEWS


Ben-Epstein (approved):

setu4993 (approved):
Curious about why we're changing it?

PR #676, opened 2023-06-15
Changed names of bucket and dataset in semseg watch

dcaustin33:
Changed name of parameters to make them more intuitive - called out by voxel. Will update the docs


dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -50,8 +50,8 @@
 class SemanticTorchLogger(TorchLogger):
     def __init__(
         self,
-        bucket_url: str,
-        dataset_path: str,
+        remote_img_location: str,
```
elboy3, line 53: IC (for good or bad) uses `imgs_remote_location` should we just use the same?

elboy3 (approved):

PR #680, opened 2023-06-20
Feature/semseg repo model support

dcaustin33:
Creates support in semantic segmentation for the repo https://github.com/qubvel/segmentation_models.pytorch


Ben-Epstein (approved):
how does this enable the other repo exactly?

dcaustin33:
> how does this enable the other repo exactly?

The hooks are altered to support the output format (tensor not dict) - as well as detaching hooks at the start to avoid embedding issues that non pytorch hub models would run into

elboy3 (approved):

PR #682, opened 2023-06-21
Ensure no labels in Prepare for IC for inference

elboy3:
Small fix for IC inference flow. This should allow us to log the dataset for IC inference. We will then have further updates to the watch function (to come) 

```python
# 🔭🌕 Galileo log: Log dataset
dq.log_image_dataset(inf_dataset, split=INF_SPLIT_NAME, inference_name=INF_NAME, imgs_remote_location="gs://galileo-public-data/CV_datasets/ImageNet10_animals_train_val/inference")
```

Screen Shot 2023-06-21 at 5 29 36 PM


setu4993 (approved):
Nice! Can we please also add a test that verifies that we raise the error you showed in the OP?

elboy3:
> Nice! Can we please also add a test that verifies that we raise the error you showed in the OP?

Hmm I'm not sure that's necessary or maybe I don't follow. This fix now prevents that error from being raised. So if you remove my fix but keep the same test, it would raise a "you can't set labels" error. But now we just ignore the labels if it's inference and so the assertion passes and we don't raise that error.

setu4993:
> This fix now prevents that error from being raised. So if you remove my fix but keep the same test, it would raise a "you can't set labels" error. But now we just ignore the labels if it's inference and so the assertion passes and we don't raise that error. 

Ah, thanks for explaining. I misunderstood the intent as to raise an error (from a different part of code than was being touched), which is why I was surprised the new test passed. 

The PR makes more sense now 🙂.

bogdan-galileo:
Thanks elboy!!

PR #683, opened 2023-06-21
Feature/semseg binary classification

dcaustin33:
Supports binary classification in dq for voxel


dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -377,16 +388,26 @@ def get_argmax_probs(
         preds = F.interpolate(preds, size=input_shape, mode="bilinear")
 
         # checks whether the model is (n, classes, w, h), or (n, w, h, classes)
-        if preds.shape[1] == self.number_classes:
+        # takes the max in case of binary classification
+        if max(preds.shape[1], 2) == self.number_classes:
             preds = preds.permute(0, 2, 3, 1)
         assert (
-            preds.shape[-1] == self.number_classes
+            max(preds.shape[3], 2) == self.number_classes
```
setu4993, line 395: We could still use the last index (`-1`), right?

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -377,16 +388,26 @@ def get_argmax_probs(
...
         logits = preds.clone()  # (bs, w, h, classes)
-        probs = (torch.nn.Softmax(dim=-1)(logits)).cpu()
+        if preds.shape[3] > 1:
+            probs = (torch.nn.Softmax(dim=-1)(logits)).cpu()
+        else:
+            probs = (torch.nn.Sigmoid()(logits)).cpu()
+
+        # expands the binary classification to a 2 channel tensor
+        if probs.shape[3] == 1:
+            probs = self.expand_binary_classification(probs)
```
setu4993, line 407: We can move this above, right?
```suggestion
        else:
            probs = (torch.nn.Sigmoid()(logits)).cpu()
            # expands the binary classification to a 2 channel tensor
            probs = self.expand_binary_classification(probs)
```

setu4993 (commented):
Useful update! Thanks, Derek!

If it's not too hard, would be nice to add a test for a dataset that does binary classification. Else, open to merging as-is.

dataquality/integrations/torch_semantic_segmentation.py
dcaustin33: We could either way I thought 3 might be more precise but either works - would love to merge so as to get it to voxel but we do need to add some semseg tests for sure

setu4993: Sounds good, thanks!

setu4993 (approved):
@dcaustin33 : Can you please bump the version in `__init__.py` before merging so that we can release with a new version?

Thanks!

dataquality/integrations/torch_semantic_segmentation.py
```diff
@@ -361,6 +361,17 @@ def calculate_mislabeled_pixels(
...
         return mislabeled_pixels
 
+    def expand_binary_classification(self, probs: torch.Tensor) -> torch.Tensor:
+        """Expands the binary classification to a 2 channel tensor
+
+        Args:
+            probs (torch.Tensor): binary classification tensor
+
+        Returns:
+            torch.Tensor: bs, 2, h, w tensor
```
anthonycorletti, line 371: Is bs batch size?

anthonycorletti (approved):

dataquality/integrations/torch_semantic_segmentation.py
dcaustin33: Yep!

PR #685, opened 2023-06-22
Feature/semseg empty case

dcaustin33:
handles the case where an image has no prediction other than background and no labels other than background


dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -207,6 +213,10 @@ def _get_data_dict(self) -> Dict:
         pred_polygons_batch, gold_polygons_batch = find_polygons_batch(
             self.pred_masks, self.gold_masks
         )
+        for i in range(len(self.image_ids)):
+            if pred_polygons_batch[i] == [] and gold_polygons_batch[i] == []:
+                pred_polygons_batch[i] = [Polygon.empty_polygon()]
```
elboy3, line 218: can you add some comments explaining what we're doing and why we do it

dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -137,9 +138,14 @@ def get_polygon_data(
         accuracies = []
         mislabeled_classes = []
         mislabeled_class_pcts = []
+
+        # create contours folder
+        os.makedirs(self.local_contours_path, exist_ok=True)
```
elboy3, line 143: why here? i think we should just do this at the beginning of watch or something

dataquality/schemas/semantic_segmentation.py
```diff
@@ -133,3 +134,21 @@ def contours_json(self) -> List:
             pixels = [pixel.deserialize_json for pixel in contour.pixels]
             contours.append(pixels)
         return contours
+
+    @staticmethod
+    def empty_polygon() -> "Polygon":
+        dep = np.random.randint(0, 100) / 100.0
```
elboy3, line 140: should probably not push it with the random dep. 

also a good docstring explaining what's going on would be good

dataquality/schemas/semantic_segmentation.py
```diff
@@ -133,3 +134,21 @@ def contours_json(self) -> List:
...
+            label_idx=-1,
+            cls_error_data=None,
+            error_type=ErrorType.none,
+            background_error_pct=0.0,
+            contours=[Contour(pixels=[Pixel(x=0, y=0)])],
+            data_error_potential=dep,
+            ghost_percentage=0.0,
+            area=1,
+            likely_mislabeled_pct=0.0,
+            class_type=ClassType.gold,
```
elboy3, line 153: if we are using pred in dataquality/loggers/model_logger/semantic_segmentation.py we should do pred here

dataquality/loggers/model_logger/semantic_segmentation.py
dcaustin33: We define local_contours_path in the logger so think we should do it here

dataquality/loggers/model_logger/semantic_segmentation.py
```diff
@@ -207,6 +211,16 @@ def _get_data_dict(self) -> Dict:
         pred_polygons_batch, gold_polygons_batch = find_polygons_batch(
             self.pred_masks, self.gold_masks
         )
+        
+        # in the case that both pred and gold polygons are empty, we need to 
+        # add an empty polygon in order to have an entry for that image in 
+        # the polygon df and thus show it in the UI
+        # therefore we add an empty polygon to the pred and have the api filter
+        # out empty polygons in the processing step
```
anthonycorletti, line 219: Confused about this comment, what's it out-ing?

dcaustin33: Basically if both a pred and gold did not have a contour it broke our run - so we add a Dummy polygon with dummy data if that happens which fixes our problem. @elboy3 will have to filter in runners those polygons so as to not send to the api our dummy polygon which right now is showing up as a null class.

elboy3 (approved):

PR #695, opened 2023-06-26
Feat/seq2seq

Ben-Epstein:
* `log_dataset` for seq2seq
* `log_model_outputs` for seq2seq
* validation
* tests


dataquality/core/log.py
```diff
@@ -403,6 +411,7 @@ def log_model_outputs(
         logits=logits.astype(np.float32) if isinstance(logits, np.ndarray) else logits,
         probs=probs.astype(np.float32) if isinstance(probs, np.ndarray) else probs,
         inference_name=inference_name,
+        labels=labels.astype(np.float32) if isinstance(labels, np.ndarray) else labels,
```
setu4993, line 414: Nice update.

dataquality/loggers/model_logger/base_model_logger.py
```diff
@@ -23,6 +23,8 @@
 
 
 class BaseGalileoModelLogger(BaseGalileoLogger):
+    log_file_ext = "hdf5"
```
setu4993, line 26: We should move this to under `__init__` instead of defining it here.

dataquality/loggers/base_logger.py
```diff
@@ -230,7 +230,7 @@ def _convert_tensor_ndarray(
                     )
             elif attr == "Labels" and config.task_type == TaskType.text_multi_label:
                 arr = np.array(arr, dtype=object)
-        return np.array(arr)
+        return np.array(arr) if not isinstance(arr, np.ndarray) else arr
```
setu4993, line 233: Nice change.

dataquality/core/log.py
```diff
@@ -388,6 +392,10 @@ def log_model_outputs(
     assert (probs is not None) or (
         logits is not None
     ), "You must provide either logits or probs"
+    # No embeddings ever provided by user in seq2seq
+    if config.task_type == TaskType.seq2seq:
+        exclude_embs = True
+        embs = None
```
setu4993, line 398: I wonder if we should start to have a task level config...

dataquality/loggers/logger_config/seq2seq.py
```diff
@@ -0,0 +1,19 @@
...
+
+from transformers import PreTrainedTokenizerFast
+
+from dataquality.loggers.logger_config.base_logger_config import BaseLoggerConfig
+
+
+class Seq2SeqLoggerConfig(BaseLoggerConfig):
+    sample_length: Dict[str, int] = {}
+    tokenizer: Optional[PreTrainedTokenizerFast] = None
+    # For each split/inference-name, store sample id -> List[tokens] for the label
```
setu4993, line 12:
```suggestion
    # For each split/inference-name, store sample id -> List[token_id] for the label
```

dataquality/core/log.py
```diff
@@ -607,3 +616,20 @@ def set_epoch_and_split(
...
+    Must be a fast tokenizer, and must support `decode`, `encode`, `encode_plus`
+    """
+    task_type = _get_task_type()
+    assert task_type == TaskType.seq2seq, "This method is only supported for seq2seq"
+    assert isinstance(
+        tokenizer, PreTrainedTokenizerFast
+    ), "Tokenizer must be an instance of PreTrainedTokenizerFast"
+    assert getattr(tokenizer, "is_fast", False), "Tokenizer must be a fast tokenizer"
+    for attr in ["encode", "decode", "encode_plus", "padding_side"]:
+        assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
```
setu4993, line 634: Good validations here.

dataquality/utils/arrow.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def save_arrow_file(location: str, file_name: str, data: Dict) -> None:
+    """
+    Helper function to save a dictionary as an hdf5 file that can be read by vaex
+    """
+    if not os.path.isdir(location):
+        with lock:
+            os.makedirs(location, exist_ok=True)
+    write_feather(pa.Table.from_pydict(data), f"{location}/{file_name}")
```
setu4993, line 17: Is there a reason we prefer feather here rather than hdf (like other tasks), given we have a bunch of numpy arrays?

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class Seq2SeqDataLogger(BaseGalileoDataLogger):
+    """Logging input data for Seq2Seq fine-tuning tasks
+
+    Logging input data for Seq2Seq requires 3 pieces of information:
+    1. tokenizer: This must be an instance of PreTrainedTokenizerFast from huggingface
+        (ie T5TokenizerFast or GPT2TokenizerFast, etc). Your tokenizer should have an
+        `.is_fast` property that returns True if it's a fast tokenizer.
+        This class must implement the `encode`, `decode`, and `encode_plus` methods
```
setu4993, line 34: I'm guessing this is because we want it to be fast, but technically could be a non-fast tokenizer too, correct?

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+from dataquality.utils.vaex import rename_df
+
+if TYPE_CHECKING:
+    from datasets import Dataset
+
+
+class Seq2SeqDataLogger(BaseGalileoDataLogger):
+    """Logging input data for Seq2Seq fine-tuning tasks
+
+    Logging input data for Seq2Seq requires 3 pieces of information:
```
setu4993, line 30:
```suggestion
    Logging input data for Seq2Seq requires 2 pieces of information:
```

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+        return str(self.split)
+
+    def validate_and_format(self) -> None:
+        super().validate_and_format()
+        label_len = len(self.labels)
+        text_len = len(self.texts)
+        id_len = len(self.ids)
+        assert id_len == text_len == label_len, (
+            "IDs, text, and labels must be the same length, got "
+            f"(ids, text, labels) ({id_len}, {text_len}, {label_len})"
```
setu4993, line 96:
```suggestion
            f"({id_len} ids, {text_len} text, {label_len} labels)"
```

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+        assert (
+            self.logger_config.tokenizer is not None
+        ), "You must set your tokenizer before logging. Use `dq.set_tokenizer`"
```
setu4993, line 100: Not formatted to 88, but the below should be fine.

```suggestion
        assert self.logger_config.tokenizer, "You must set your tokenizer before logging. Use `dq.set_tokenizer`"
```

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+        encoded_data = self.logger_config.tokenizer(
+            self.labels, return_offsets_mapping=True
+        )
```
setu4993, line 103: Why do we need to do this instead of relying on the dataloader to give this to us as a part of the batch?

setu4993 (commented):
I still need to look through the model logger, utils and tests, but it is looking fairly good to me. Left a few questions so that I can understand this better.

docs/cv/coco_hf_datasets.py
```diff
@@ -10,13 +10,13 @@
...
                  dataset_path: str,
-                 relative_img_path: Optional[str], 
+                 relative_img_path: Optional[str],
                  relative_mask_path: Optional[str],
-                 mask_transform: transforms=None, 
-                 img_transform: transforms=None, 
-                 size: int=1024,
+                 mask_transform: transforms = None,
+                 img_transform: transforms = None,
+                 size: int = 1024,
```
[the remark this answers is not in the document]
Ben-Epstein: no im not sure what happened here . I think linting must have been missed

dataquality/loggers/data_logger/seq2seq.py
Ben-Epstein: @jonathangomesselman said that wouldn't be easy/always possible

dataquality/loggers/data_logger/seq2seq.py
Ben-Epstein: @setu4993 nope, the non-fast tokenizer doesn't implement `encode_plus`

dataquality/loggers/model_logger/base_model_logger.py
```diff
@@ -32,6 +34,7 @@ def __init__(
         split: str = "",
         epoch: Optional[int] = None,
         inference_name: Optional[str] = None,
+        labels: Optional[np.ndarray] = None,
```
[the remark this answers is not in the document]
Ben-Epstein: we try to not rely on kwargs because it makes our signatures less clean. But also we'll be removing `labels` anyway so Ill update

dataquality/loggers/model_logger/base_model_logger.py
Ben-Epstein: hm i don't agree. This is an attribute of the Class, not an instance of the class.

dataquality/core/log.py
Ben-Epstein: we do! `logger_config` is that task level config

dataquality/utils/arrow.py
Ben-Epstein: we dont in seq2seq. I would prefer hdf5, but the things we save have to be pyarrow arrays because each row has a differnt sized list (and padding would be really expensive on size). So we need pyarrow arrays, hence arrow files

dataquality/loggers/data_logger/seq2seq.py
setu4993: Oh interesting.

dataquality/loggers/model_logger/text_multi_label.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         split: str = "",
         epoch: Optional[int] = None,
         inference_name: Optional[str] = None,
+        labels: Optional[np.ndarray] = None,
```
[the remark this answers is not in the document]
Ben-Epstein: because we added an optional field to `log_model_outputs` at the top level. So the `seq2seqModelLogger` needed a new `__init__` param. But in order to maintain OOP all parent (and subsequent) children classes need it as well

dataquality/loggers/model_logger/base_model_logger.py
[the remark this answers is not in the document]
Ben-Epstein: Yea we try to avoid kwargs here because this is a user facing API and its harder to interpret. But i understand the argument both sides. I think for now I'd prefer to keep it like this. If it grows unruly we can always move over

dataquality/loggers/data_logger/seq2seq.py
Ben-Epstein: ill update it locally 👍

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -0,0 +1,142 @@
...
+            logits=logits,
+            ids=ids,
+            split=split,
+            epoch=epoch,
+            inference_name=inference_name,
+            labels=labels,
+        )
+        self.token_dep = pa.array([])
+        self.token_gold_probs = pa.array([])
+        self.labels = labels
```
setu4993, line 41: Do we need to do this even though we're passing it above to the parent init?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -0,0 +1,142 @@
...
+            "Must pass in a valid batch with equal id and logit length, got "
+            f"id: {len(self.ids)},logits: {len(self.logits)}"
+        )
+        if self.labels is not None:
+            assert len(self.labels) == len(self.ids), "TODO: Must be same len message"
+
+        # Ground truth probs, including the padding for ignored labels
+        # TODO: This is incredibly slow. This is what needs to be optimized. Can we
+        #  potentially do this on the GPU with torch? And dont convert to a np array
+        probs = self.convert_logits_to_probs(self.logits)
```
setu4993, line 65: Seems doable on a GPU.

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -0,0 +1,142 @@
...
+            self.logger_config.tokenizer is not None
+        ), "Must set your tokenizer. Use `dq.set_tokenizer`"
+        labels = self.logger_config.id_to_tokens[self.token_map_key][sample_id]
+        if self.logger_config.tokenizer.padding_side == "left":
+            probs = sample_probs[-len(labels) :]
+        else:
+            probs = sample_probs[: len(labels)]
+        gold_probs = probs[np.arange(len(labels)), labels]
+        probs_copy = probs.copy()
+        probs_copy[np.arange(len(labels)), labels] = 0
```
setu4993, line 83: Nit: We should likely set this to a nan.

dataquality/utils/seq2seq.py
```diff
@@ -0,0 +1,268 @@
...
+    We take overlapping ranges and gaps between ranges, and fill them in contiguously
+
+    ex:
+        offset_mapping = [(0, 1), (0, 20), (22, 23), (0, 0)]
+
+        is rolled into
+            [
+                {'offsets': (0, 1), 'token_positions': {0, 1}},
+                {'offsets': (1, 20), 'token_positions': {1}},
+                {'offsets': (20, 22), 'token_positions': {}},
```
setu4993, line 142: This is when it's whitespace?

dataquality/utils/seq2seq.py
```diff
@@ -0,0 +1,268 @@
...
+        span_offsets.append(cur_offset)
+        span_mapping[cur_offset].add(cur_tok_pos)
+    # Gap in tokens (like a space). fill the gap with no tokens
+    elif last_end < cur_start:
+        # Fill the empty space with nothing
+        span_offsets.append((last_end, cur_start))
+        span_mapping[(last_end, cur_start)] = set()
+        # Then add the current span
+        span_offsets.append(cur_offset)
+        span_mapping[cur_offset].add(cur_tok_pos)
```
setu4993, line 107: I don't particularly like that we're modifying an object that is created in the parent fn...

dataquality/utils/seq2seq.py
```diff
@@ -0,0 +1,268 @@
...
+from typing import Dict, List, Set, Tuple
+
+import numpy as np
+import pyarrow as pa
+from tqdm.auto import tqdm
+
+from dataquality.schemas.seq2seq import AlignedTokenData
+
+
+def _handle_overlapping_offsets(
```
setu4993, line 11: This fn is pretty deep. Nice work optimizing + adding comments.

tests/loggers/test_seq2seq.py
```diff
@@ -0,0 +1,157 @@
...
+    token_gold_probs = output_data["token_gold_probs"].tolist()
+    token_deps = output_data["token_deps"].tolist()
+
+    for token_labels, token_dep, token_gold_prob in zip(
+        tokenized_labels, token_deps, token_gold_probs
+    ):
+        assert len(token_labels) == len(token_dep) == len(token_gold_prob)
+        assert -100 not in token_gold_prob
+        for dep in token_dep:
+            assert 0 <= dep <= 1
```
setu4993, line 157: Good validations.

setu4993 (approved):
🚀

dataquality/utils/seq2seq.py
Ben-Epstein: spaces that arent taken into account by the tokenizer. There were a bunch of examples of those

dataquality/utils/seq2seq.py
Ben-Epstein: yea it was such a nested function I couldn't find a cleaner way to do it

dataquality/core/log.py
Ben-Epstein: potentially, but there are more checks around this below. Maybe we can clean it all up in a follow-up pr

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -0,0 +1,188 @@
...
+        self.logger_config.id_to_tokens[self.token_map_key].update(id_to_tokens)
+
+    def _get_input_df(self) -> DataFrame:
+        return vaex.from_dict(
+            {
+                C.id.value: self.ids,
+                C.text.value: self.texts,
+                C.label.value: self.labels,
+                # TODO: Do we need to save this to the server?
+                C.tokenized_label.value: pa.array(self.tokenized_labels),
```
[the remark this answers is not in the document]
Ben-Epstein: the tokenized index

dataquality/loggers/model_logger/seq2seq.py
Ben-Epstein: yea i could have done something like
```
if labels is not None:
    self.labels = labels
```

in the parent but i ignored it for now

dataquality/utils/seq2seq.py
setu4993: Yup, makes sense. That's what I expected.

dataquality/loggers/data_logger/seq2seq.py
jonathangomesselman: Two things here:
1) We actually do need to tokenizer this ourselves in order to get the `offsets_mapping` which is key for aligning the characters + tokens in the UI
2) For the model output side, I was not sure how easy it would be through an integration to track the data loader / capture the labels field from the HuggingFace Trainer. It is certainly there since it is used to compute the loss, but from talking with Franz it seemed like it may require some patching or similar techniques. For this reason, it felt easiest to start with just referencing the labels internally.

PR #700, opened 2023-06-29
downcast before saving batch

Ben-Epstein:
we've done this in a few places before uploading to Galileo, but we still write the batches to disk. This can really add up (f64 embeddings are huge).

Safely always downcast before writing to disk


setu4993 (approved):
💯

elboy3 (approved):
👨‍🍳 💋

PR #702, opened 2023-07-05
Add early stopping optional for auto

elboy3:
Comcast feedback was that they don't always want early stopping

Also removing open console url for reddit request
https://app.shortcut.com/galileo/story/6400/dq-remove-console-url


franz101 (commented):
Looks good to me, can we test it in a colab?

franz101:
Not passing the callback is the right approach. One separate thing is to watch out for the flag load_best_model_at_the_end

dataquality/core/finish.py
```diff
@@ -87,7 +87,6 @@ def finish(
             "Don't close laptop or terminate shell."
         )
         wait_for_run()
-        open_console_url(res["link"])
```
setu4993, line 90: Can we make this dependent on a flag instead of always?

elboy3: Yeah we could - do you like the automatic console url? 

we agreed in slack to fully remove it https://galileoteamworkspace.slack.com/archives/C03UZ2K3D2L/p1688154343802109

setu4993: I do, but if that's what we've agreed to, sure, let's do it.

setu4993 (approved):

PR #703, opened 2023-07-06
Retry upload 3 times in case of failure

franz101:
Adds 3 tries for retrying to upload in case of failure.

- Does not depend on external libraries


pyproject.toml
```diff
@@ -67,7 +67,7 @@ test = [
     "coverage[toml]>=7.0.5",
     "pytest-cov>=4.0.0",
     "scikit-learn>=1.0",
-    "tensorflow>=2.9.1",
+    "tensorflow>=2.9.1,<2.13.0",
```
elboy3, line 70: so folks can never use a higher version of tensorflow?

dataquality/clients/api.py
```diff
@@ -82,15 +84,23 @@ def make_request(
...
+                        params=params,
+                        headers=header,
+                        data=data,
+                        timeout=timeout,
+                        files=files,
+                    )
+        except RetryError as e:
+            raise GalileoException(
+                f"Failed to make request after {max_retries} attempts. " f"Error: {e}"
+            )
```
elboy3, line 103: let's chat about this. i would just use the tenacity decorator. 1 line instead of 17

pyproject.toml
franz101: needs to be fixed sperately

bogdan-galileo: This req is in the `test` extras. But if they do happen to use tf >= 2.13.0 our method https://github.com/rungalileo/dataquality/blob/9b8af3636decb6bb5a7828df7fe6ac7e755188f3/dataquality/utils/keras.py#L46C1-L54C6 will fail and it seems to be used everywhere.

It's a prob in TF so I opened an issue on TF https://github.com/tensorflow/tensorflow/issues/61204#issuecomment-1624338234

bogdan-galileo (approved):
LGTM !

elboy3 (approved):

PR #704, opened 2023-07-06
Feat/cv save local paths

bogdan-galileo:
Main goal is to start logging the local paths to images for IC (necessary for the smart features to run: blurriness, bad exposure, etc), see ticket https://app.shortcut.com/galileo/story/6278/cv-allow-logging-for-local-images

- added a non_metadata attribute to the TC logger, which works and logs data like metadata, but doesn't show up in the UI
- cleaned up and simplified the interface to log_image_dataset + updated docstring
- added lots of tests for log_image_dataset covering all scenarios

Probably require minor version bump as it's not backwards compatible


dataquality/loggers/data_logger/base_data_logger.py
```diff
@@ -73,10 +74,13 @@ class BaseGalileoDataLogger(BaseGalileoLogger):
...
     INPUT_DATA_FILE_EXT = "arrow"
 
-    def __init__(self, meta: Optional[MetasType] = None) -> None:
+    def __init__(
+        self, meta: Optional[MetasType] = None, non_meta: Optional[MetasType] = None
+    ) -> None:
         super().__init__()
         self.meta: Dict = meta or {}
         self.log_export_progress = True
+        self.non_meta: Dict = non_meta or {}  # Extra data to log (that is not metadata)
```
anthonycorletti, line 83: wdyt about naming this `extra_data_cols` or `extra_cols`?

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -77,67 +78,95 @@ def log_image_dataset(
...
-            )
-        return os.path.isfile(dataset[imgs_location_colname].iloc[0])
+    def _has_remote_images(self, dataset: DataSet) -> bool:
+        """Check if the dataset contains a column containing remote images"""
+        if isinstance(dataset, pd.DataFrame):
+            columns = dataset.columns
+        elif self.is_hf_dataset(dataset):
+            columns = dataset.column_names  # type: ignore # noqa: F821
+
+        # TODO: should we also check that we can reach one of the images ?
```
anthonycorletti, line 145: interesting! i think we should assume that it will be reachable as the server is the responsible part for delivering the images to the ui. if something is wrong we know there's one place that has to reach the images.

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -220,20 +269,30 @@ def _prepare_df_from_ImageFolder(
...
             }
             df["label"] = df.label.map(label_idx_to_label)
 
         df = df.reset_index().rename(columns={"index": "id"})
 
-        # Replace the paths with the remote one, if a remote location is specified
+        # Also add remote paths, if a remote location is specified
         if imgs_remote_location is not None:
-            df["text"] = df["text"].str.replace(dataset.root, imgs_remote_location)
+            self.imgs_remote_colname = "text_original"
```
anthonycorletti, line 292: wdyt about pulling `text_original` into a const?

dataquality/loggers/data_logger/text_classification.py
```diff
@@ -156,6 +157,7 @@ def log_data_samples(
         self.labels = [str(i) for i in labels] if labels is not None else []
         self.inference_name = inference_name
         self.meta: Union[MetasType, Dict] = meta or {}
+        self.non_meta: Union[MetasType, Dict] = non_meta or {}
```
anthonycorletti, line 160: wdyt about renaming to `extra_cols`?

dataquality/loggers/data_logger/base_data_logger.py
bogdan-galileo: Haha! I think I had exactly `extra_cols` at the very beginning and then renamed for some reason. Renaming to `extra_cols`!

dataquality/loggers/data_logger/text_classification.py
bogdan-galileo: done!

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: yeah, where would I store it ? I can look a bit. The thought here is that stuff breaks if they already have a column called "text" since we use that internally to store our paths (we get a df that has two "text" columns and looks weird)

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: Oh yeah I agree, I meant it as: should we check that they didn't mess up the URL ?

dataquality/loggers/data_logger/image_classification.py
anthonycorletti: a global var in this module would do. unsure if it will be used in other places.

bogdan-galileo: added!

anthonycorletti (approved):
:shipit:

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -77,67 +80,93 @@ def log_image_dataset(
         parallel: bool = False,
     ) -> Any:
         """
-        For docstring see top level method located in core/log.py
+        For main docstring see top level method located in core/log.py.
+
+        In addition we set the variables self.imgs_local_colname and
+        self.imgs_remote_location.
```
elboy3, line 86: i think in the docstring its better to include the **why** and not just the **what**

What do we have to set these?

dataquality/loggers/data_logger/image_classification.py
```diff
@@ -77,67 +80,93 @@ def log_image_dataset(
...
         """
         if type(dataset).__name__ == "ImageFolder":
+            # imgs_local will be ignored as it is not necessary
```
elboy3, line 89: why isn't this var necessary here?

dataquality/loggers/data_logger/text_classification.py
```diff
@@ -148,6 +149,8 @@ def log_data_samples(
             data is required. Can be set here or via dq.set_split
         :param meta: Dict[str, List[str, int, float]]. Metadata for each text sample
             Format is the {"metadata_field_name": [metdata value per sample]}
+        :param extra_cols: Dict[str, List[str, int, float]]. Extra columns to log that
+            is not Metadata for each text sample. Same format as Metadata
```
elboy3, line 153: can you include in the docstring what would be a good example of an extra col?

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: Removed that comment since these are just internal vars. I initially thought these are vars that persist when you call get_data_logger

dataquality/loggers/data_logger/image_classification.py
bogdan-galileo: added comment

elboy3 (approved):
🚀

PR #708, opened 2023-07-07
fix: Move version validation to `init`, log warning instead of raising exception

setu4993:
Add test for #707.

### DIFF
--- a/tests/utils/test_version.py
+++ b/tests/utils/test_version.py
@@ -1,16 +1,9 @@
 from unittest import mock
 from unittest.mock import MagicMock
 
-import pytest
-
 from dataquality import __version__
-from dataquality.exceptions import GalileoException
 from dataquality.utils import version
-from tests.test_utils.mock_request import (
-    MockResponse,
-    mocked_healthcheck_request,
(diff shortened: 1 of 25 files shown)


tests/utils/test_version.py
```diff
@@ -35,3 +36,9 @@ def test_version_check_fail(mock_get_healthcheck: MagicMock) -> None:
 def test_version_check_pass(mock_get_healthcheck: MagicMock) -> None:
     version._version_check()
     assert True
+
+
+@mock.patch("requests.get", return_value=MockResponse({"api_version": __version__.replace("v", "")}, 200))
+def test_version_check_pass_without_v(mock_get_healthcheck: MagicMock) -> None:
+    version._version_check()
+    assert True
```
elboy3, line 44: don't need the assert

elboy3: ahh i see the other tests have it 🤷‍♀️

elboy3 (approved):

tests/utils/test_version.py
setu4993: Lol, I just copied that over. I'll remove it.

PR #712, opened 2023-07-10
Version bump for robustification

franz101:


dataquality/__init__.py
```diff
@@ -31,7 +31,7 @@
 """
 
 
-__version__ = "0.9.5"
+__version__ = "v0.9.6"
```
setu4993, line 34:
```suggestion
__version__ = "0.9.6"
```

setu4993 (changes requested):
We dropped the `v` previously, let's remove it.

setu4993 (approved):

PR #715, opened 2023-07-18
Chore: Documentation and input / output cols

elboy3:
Small cleanup leading up to S2S sprint during onsite week

### DIFF
--- a/dataquality/loggers/model_logger/seq2seq.py
+++ b/dataquality/loggers/model_logger/seq2seq.py
@@ -5,7 +5,7 @@ import pyarrow as pa
 
 from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
 from dataquality.loggers.model_logger.base_model_logger import BaseGalileoModelLogger
-from dataquality.schemas.seq2seq import Seq2SeqCols as C
+from dataquality.schemas.seq2seq import Seq2SeqOutputCols as C
 from dataquality.schemas.split import Split
 from dataquality.utils.arrow import save_arrow_file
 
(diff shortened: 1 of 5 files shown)


dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -135,7 +154,7 @@ def _get_data_dict(self) -> Dict:
             C.epoch.value: [self.epoch] * batch_size,
         }
         if self.split == Split.inference:
-            data["inference_name"] = [self.inference_name] * batch_size
+            data[C.inference_name.value] = [self.inference_name] * batch_size
```
setu4993, line 157: Do we need a `.value` here? I think it works simply, too?
```suggestion
            data[C.inference_name] = [self.inference_name] * batch_size
```

dataquality/schemas/seq2seq.py
```diff
@@ -3,18 +3,23 @@
...
+
+
+class Seq2SeqOutputCols(str, Enum):
+    id = "id"
     token_deps = "token_deps"
     token_gold_probs = "token_gold_probs"
     # Mypy complained about split as an attribute, so we use `split_`
     split_ = "split"
     epoch = "epoch"
+    inference_name = "inference_name"
```
setu4993, line 22: Why `inference_name`?

setu4993 (approved):

dataquality/schemas/seq2seq.py
elboy3: For line 157 in dataquality/loggers/model_logger/seq2seq.py

dataquality/loggers/model_logger/seq2seq.py
elboy3: I did value for consistency with the above 5 lines, but I can do a test removing .value from all of them and push that up

elboy3: Yes we do! This creates a dict that later gets cast to a vaex DataFrame, which can't have enum as the keys to the dict 

The following errors:
```python
from enum import Enum
import vaex

class MyType(str, Enum):
    a = "a"
    b = "b"
    c = "c"

df = vaex.from_dict(
    {
        MyType.a: [1, 2, 3],
        MyType.b: [4, 5, 6],
        MyType.c: [7, 8, 9],
    }
)
```

setu4993: TIL!

dataquality/schemas/seq2seq.py
setu4993: My question was why call it inference _name_? _Name_ as a column name is too generic and doesn't feel like that's what we should be getting?

anthonycorletti (approved):

PR #716, opened 2023-07-19
Fix/seq2seq finish

elboy3:
https://app.shortcut.com/galileo/story/6733/dq-seq2seq-finish-and-upload-support

Update the `.finish()` flow to combine and upload the data.arrow file for Seq2Seq

We pin `transformers==4.30.0` due to a new error introduced with 4.31.0 here: https://github.com/rungalileo/dataquality/actions/runs/5603321961/jobs/10249775309?pr=716

Franz is looking into a fix and then we will unpin transformers

### DIFF
--- a/dataquality/core/log.py
+++ b/dataquality/core/log.py
@@ -633,3 +633,5 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
     seq2seq_logger_config.tokenizer = tokenizer
+    # Seq2Seq doesn't have labels but we need to set this to avoid validation errors
+    seq2seq_logger_config.labels = []
--- a/dataquality/loggers/data_logger/seq2seq.py
+++ b/dataquality/loggers/data_logger/seq2seq.py
@@ -191,7 +191,6 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
 
     @classmethod
     def _get_prob_cols(cls) -> List[str]:
-        # return ["id", "token_deps", "token_gold_probs"]
         return ["id"]
 
     @classmethod
--- a/dataquality/loggers/data_logger/seq2seq.py
+++ b/dataquality/loggers/data_logger/seq2seq.py
@@ -13,6 +13,7 @@ from dataquality.loggers.data_logger.base_data_logger import (
     MetasType,
 )
 from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
+from dataquality.schemas.dataframe import BaseLoggerDataFrames
 from dataquality.schemas.seq2seq import Seq2SeqInputCols as C
 from dataquality.schemas.split import Split
 from dataquality.utils.seq2seq import (
@@ -67,6 +68,7 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
 
     __logger_name__ = "seq2seq"
     logger_config = seq2seq_logger_config
+    DATA_FOLDER_EXTENSION = {"emb": "hdf5", "prob": "hdf5", "data": "arrow"}
 
     def __init__(self, meta: Optional[MetasType] = None) -> None:
         super().__init__(meta)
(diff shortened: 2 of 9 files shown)


dataquality/core/log.py
```diff
@@ -633,3 +633,5 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
     seq2seq_logger_config.tokenizer = tokenizer
+    # Seq2Seq doesn't have labels but we need to set this to avoid validation errors
+    seq2seq_logger_config.labels = []
```
setu4993, line 637: We do have labels, though, right? Are we maybe saying it's not expected to be set at this point?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "cachetools>=4.2.4",
     "importlib-metadata<6.0.1",
     "datasets>=2.6",
-    "transformers>=4.17.0",
+    "transformers>=4.17.0,<4.31.0",
```
setu4993, line 32: 🤔

dataquality/utils/file.py
```diff
@@ -4,10 +4,29 @@
...
+    """
+    files = os.listdir(dir_)
+    if len(files) == 0:
+        raise GalileoException(f"No files found in {dir_}")
+
+    extensions = set([get_file_extension(f) for f in files])
+    if len(extensions) > 1:
+        raise GalileoException(
+            f"Multiple file extensions found in {dir_}: {extensions}"
+        )
```
setu4993, line 25: Feels a bit fragile, honestly.

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -67,6 +68,7 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
 
     __logger_name__ = "seq2seq"
     logger_config = seq2seq_logger_config
+    DATA_FOLDER_EXTENSION = {"emb": "hdf5", "prob": "hdf5", "data": "arrow"}
```
setu4993, line 71: Where does this get used?

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -186,3 +188,35 @@ def get_valid_attributes() -> List[str]:
...
         return list(map(lambda x: x.value, C))
+
+    @classmethod
+    def _get_prob_cols(cls) -> List[str]:
+        return ["id"]
+
+    @classmethod
+    def separate_dataframe(
+        cls, df: DataFrame, prob_only: bool = True, split: Optional[str] = None
+    ) -> BaseLoggerDataFrames:
```
setu4993, line 199: Nice. I like that we split these up.

setu4993 (commented):
Good update. Left a few questions to clarify a few things.

dataquality/utils/file.py
elboy3: Any suggestions? the flow currently expects all files to be hdf5 so this at least now allows for hdf5 and arrow

dataquality/loggers/data_logger/seq2seq.py
elboy3: this is the file format we save for the files uploaded to teh root bucket in minio

pyproject.toml
elboy3: see PR description!

dataquality/core/log.py
elboy3: nope, not dataset ground truth labels like we have in TC or NER. we have a `target` column but that's different than this

pyproject.toml
setu4993: Thanks, just saw :).

dataquality/core/log.py
setu4993: Ah okay, we call that `target` here, makes sense.

dataquality/utils/file.py
setu4993: Yeah, I think that's fair. It just feels a bit risky since we could populate files with internal code that maybe doesn't conform to this. But it's possible we will avoid that and likely never hit that from `dq`.

setu4993 (approved):

dataquality/utils/file.py
elboy3: i think it's a very specific helper we need in a very specific use case, which is that we've saved a bunch of batches to a folder (as hdf5 or arrow files) and now need to combine the batches.

so for this specific helper i do want it to alert if there are multiple types of files in the folder. 

would it help if i update the fn name or the docstring?

setu4993: Nah, I think it's good for now. I don't expect us to hit this pathway much, so it feels fine.

PR #717, opened 2023-07-20
Fixed transformers move to accelerate

franz101:
Changed the ids to a queue


tests/conftest.py
```diff
@@ -16,6 +17,10 @@
 from dataquality.utils.dq_logger import DQ_LOG_FILE_HOME
 from tests.test_utils.mock_request import MockResponse
 
+try:
+    torch.set_default_device("cpu")
+except AttributeError:
+    print("Torch default device not set to CPU")
```
elboy3, line 23: print won't be printed by default. maybe use GalileoWarn?

tests/integrations/hf/test_text_classification_hf.py
```diff
@@ -67,7 +70,7 @@ def compute_metrics(eval_pred):
 
 # Training arguments and training part
 metric_name = "accuracy"
-batch_size = 4
+batch_size = 3
```
elboy3, line 73: why?

dataquality/utils/torch.py
```diff
@@ -26,11 +26,36 @@
 from dataquality.utils.patcher import Borg, Patch, PatchManager
 
 
+@dataclass
+class ModelOutputsStore:
+    embs: Optional[Tensor] = None
+    logits: Optional[Any] = None
```
elboy3, line 32: why any?

dataquality/utils/torch.py
```diff
@@ -26,11 +26,36 @@
 from dataquality.utils.patcher import Borg, Patch, PatchManager
 
 
+@dataclass
+class ModelOutputsStore:
+    embs: Optional[Tensor] = None
```
elboy3, line 31: what if they use tensorflow?

elboy3 (approved):
had a bunch of questions. but also looks good to me. are we sure the ids align correctly here? also test to make sure it works with transformers 4.30.0

dataquality/utils/torch.py
franz101: then it would be in tensorflow/utils

dataquality/utils/torch.py
franz101: Some weird dependencies in semseg provide a type error as there it could be also a tuple

tests/integrations/hf/test_text_classification_hf.py
franz101: good catch, can be reverted

tests/conftest.py
franz101: Agree

dataquality/__init__.py
```diff
@@ -31,7 +31,7 @@
 """
 
 
-__version__ = "0.9.7"
+__version__ = "v0.9.8"
```
setu4993, line 34: Let's drop the `v` please.

tests/conftest.py
```diff
@@ -33,12 +39,12 @@
...
-        LOCAL_MODEL_PATH, device_map="cpu"
-    )
+        LOCAL_MODEL_PATH  # , device_map="cpu"
+    ).to("cpu")
 except Exception:
-    model = AutoModelForSequenceClassification.from_pretrained(HF_TEST_BERT_PATH).to(
-        "cpu"
-    )
+    model = AutoModelForSequenceClassification.from_pretrained(
+        HF_TEST_BERT_PATH  # , device_map="cpu"
```
setu4993, line 46:
```suggestion
        HF_TEST_BERT_PATH
```

dataquality/integrations/torch.py
```diff
@@ -150,21 +150,22 @@ def _on_step_end(self) -> None:
         extracted in the hooks and we need to log them in the on_step_end
         method.
         """
+
         model_outputs_store = self.torch_helper_data.model_outputs_store
         # Workaround for multiprocessing
-        if model_outputs_store.get("ids") is None and len(
+        if model_outputs_store.ids is None and len(
```
setu4993, line 156: Is the type of `model_outputs_store` changing? This feels a bit risky if not.

dataquality/utils/torch.py
setu4993: Could we use a `Union` of all of the possible types?

dataquality/integrations/torch.py
setu4993: Never mind, I see the type update below.

dataquality/utils/torch.py
```diff
@@ -26,11 +26,36 @@
...
+        """Returns the class as a dictionary."""
+        return {"embs": self.embs, "logits": self.logits, "ids": self.ids}
+
+
 @dataclass
 class TorchHelper:
     model: Optional[Any] = None
     hook_manager: Optional[Any] = None
-    model_outputs_store: Dict[str, Any] = field(default_factory=dict)
+    model_outputs_store: ModelOutputsStore = field(default_factory=ModelOutputsStore)
```
setu4993, line 58: Good change.

pyproject.toml
```diff
@@ -67,7 +67,7 @@ test = [
     "coverage[toml]>=7.0.5",
     "pytest-cov>=4.0.0",
     "scikit-learn>=1.0",
-    "tensorflow>=2.9.1,<2.13.0",
+    "tensorflow>=2.9.1,<2.13.0", # related to https://github.com/tensorflow/tensorflow/issues/61204
```
setu4993, line 70: Heh.

tests/integrations/hf/test_text_classification_hf.py
```diff
@@ -33,7 +33,10 @@
 
 def preprocess_function(examples, tokenizer):
     return tokenizer(
-        examples["text"], padding="max_length", max_length=201, truncation=True
+        examples["text"],
+        padding="max_length",
+        max_length=201,
+        truncation=True,
     )
```
setu4993, line 40: Hmm, why is this changing?

setu4993 (approved):
Left a few tiny notes, but let's please ensure we are **not** using `v` in the version number before we merge.

dataquality/__init__.py
franz101: ah was due inv version bump. fixed both parts

pyproject.toml
franz101: currently the upper bound is pinned of tensor flow, the reason is the related issue. can remove it but just in case someone wonders.

tests/integrations/hf/test_text_classification_hf.py
franz101: only formatting

PR #718, opened 2023-07-21
Chore: Rename text and label cols to input and target output

elboy3:
In this PR we:
- rename text -> input
- rename label -> target_output
- add sample level DEP

This has been successfully tested in https://deploy-preview-2770--galileo-ui.netlify.app/insights?dataframeColumns=data_error_potential%3B%26error_types%3B%26generated_output%3B%26input%3B%26target_output%3B%26token_deps%3B%26token_label_offsets%3B%26token_label_positions&projectId=bd7e3780-9f8b-4219-926e-7d3f8021410a&runId=15ae4835-8763-4b09-92b3-9db21f5cbe92&taskType=8


dcaustin33 (approved):

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -219,4 +219,6 @@ def separate_dataframe(
 
         emb = df_copy[emb_cols]
         data_df = df_copy[other_cols]
+        data_df.rename("text", "input")
+        data_df.rename("label", "target_output")
```
setu4993, line 223: Why rename here instead of when we create `data_df`?

setu4993 (approved):

dataquality/loggers/data_logger/seq2seq.py
elboy3: Maybe could have. due to the inheritance from base and the assumptions of text and label in the DF for all modalities, there might be some places the code assumes "text" in DF, so I decided to wait until the very end to rename. 

I was going to come back to this after demo and clean it up though which was going to include looking into breaking up that assumption about text and label and replacing it with more Seq2Seq appropriate jargon

PR #720, opened 2023-07-24
chore: Add py37 warning

setu4993:
During the import of `dataquality`, start warning users that we're going to be dropping Python 3.7 support soon.

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -35,9 +35,11 @@ __version__ = "0.9.8"
 
 import sys
 from typing import Any, List, Optional
+from warnings import warn
 
 import dataquality.core._config
 import dataquality.integrations
+from dataquality.exceptions import GalileoWarning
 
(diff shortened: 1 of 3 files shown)


elboy3 (commented):
Why in the Analytics init? What not in `dq.init` or in the `__init__.py` file when you import?

setu4993:
@elboy3 :

> Why in the Analytics init? What not in `dq.init` or in the `__init__.py` file when you import?

I put it here because that's where we were logging the version to Amplitude for. I personally prefer putting it in `__init__.py` but wasn't sure if that'd be okay. Can move it there.

elboy3:
@setu4993 I think i'd prefer that, since we might remove amplitude in the not so far future as well

setu4993:
@elboy3 :

> @setu4993 I think i'd prefer that, since we might remove amplitude in the not so far future as well

Fair enough, done.

elboy3 (approved):

PR #729, opened 2023-07-28
Chore: Remove cloud inference restriction

elboy3:
https://app.shortcut.com/galileo/story/6866/dq-remove-cloud-inference-restriction


anthonycorletti (approved):
release the inference hounds!

PR #731, opened 2023-08-01
Pin ipywidgets to avoid iprogress bar import issue

elboy3:
Use the latest version of [ipywidgets](https://pypi.org/project/ipywidgets/)

https://app.shortcut.com/galileo/story/6871/dq-pyproject-progress-bar-dependency#activity-6872

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -31,7 +31,7 @@ If you want to train without a model, you can use the auto framework:
 """
 
 
-__version__ = "0.10.2"
+__version__ = "0.10.3"
 
 import sys
 from typing import Any, List, Optional
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dependencies = [
     "tenacity>=8.1.0",
     "evaluate",
     "accelerate",
+    "ipywidgets>=8.1.0",
 ]
 [[project.authors]]
 name = "Galileo Technologies, Inc."


setu4993 (approved):
Are we sure we want to install `ipywidgets` for everyone, even in non-Jupyter environments?

elboy3:
@setu4993 it's a good point I was thinking about this. What else can we do? Imagine a user just goes to a jupyter notebook and does `! pip install dataquality` and then dq.auto().. we can't be throwing an error there. 

I feel like it's worth it to have all users have ipywidgets to prevent this case since it's such a small package. But lmk if there's another approach we can take

setu4993:
> @setu4993 it's a good point I was thinking about this. What else can we do? Imagine a user just goes to a jupyter notebook and does `! pip install dataquality` and then dq.auto().. we can't be throwing an error there.

Agreed.

> I feel like it's worth it to have all users have ipywidgets to prevent this case since it's such a small package. But lmk if there's another approach we can take

`ipywidgets` requires [a few packages](https://github.com/jupyter-widgets/ipywidgets/blob/main/python/ipywidgets/setup.cfg#L36-L41), so it's not really small.

Ideally, we'd distribute it as an extra, but given  `dq` runs mostly in Jupyter notebooks / training envs, where this is likely to already exist, it feels okay.

Ideally, we'd distribute it as an extra, but since we don't this seems fine.

PR #734, opened 2023-08-09
Feature/text generation seq2seq

elboy3:
>
>

***Demonstration***

### DIFF
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -1,11 +1,16 @@
 from transformers import GenerationConfig, PreTrainedModel
 
-import dataquality as dq
+from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
+from dataquality.schemas.split import Split
 from dataquality.schemas.task_type import TaskType
 from dataquality.utils.helpers import get_task_type
 
 
-def watch(model: PreTrainedModel, generation_config: GenerationConfig) -> None:
+def watch(
+    model: PreTrainedModel,
+    generation_config: GenerationConfig,
+    generate_training_data: bool = False,
+) -> None:
     """Seq2seq only. Log model generations for your run
 
     Iterates over a given dataset and logs the generations for each sample.
@@ -18,3 +23,11 @@ def watch(model: PreTrainedModel, generation_config: GenerationConfig) -> None:
         model, PreTrainedModel
     ), "model must be an instance of transformers PreTrainedModel"
     assert model.can_generate(), "model must contain a `generate` method for seq2seq"
+
+    seq2seq_logger_config.model = model
+    seq2seq_logger_config.generation_config = generation_config
+
+    generation_splits = {Split.validation, Split.test}
+    if generate_training_data:
+        generation_splits.add(Split.training)
+    seq2seq_logger_config.generation_splits = generation_splits
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -1,5 +1,6 @@
 from transformers import GenerationConfig, PreTrainedModel
 
+import dataquality as dq
 from dataquality.schemas.task_type import TaskType
 from dataquality.utils.helpers import get_task_type
 
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -1,0 +1,19 @@
+from transformers import GenerationConfig, PreTrainedModel
+
+from dataquality.schemas.task_type import TaskType
+from dataquality.utils.helpers import get_task_type
+
+
+def watch(model: PreTrainedModel, generation_config: GenerationConfig) -> None:
+    """Seq2seq only. Log model generations for your run
+
+    Iterates over a given dataset and logs the generations for each sample.
+    `model` must be an instance of transformers PreTrainedModel and have a `generate`
+      method.
+    """
+    task_type = get_task_type()
+    assert task_type == TaskType.seq2seq, "This method is only supported for seq2seq"
+    assert isinstance(
+        model, PreTrainedModel
+    ), "model must be an instance of transformers PreTrainedModel"
+    assert model.can_generate(), "model must contain a `generate` method for seq2seq"
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -3,7 +3,7 @@ from transformers import GenerationConfig, PreTrainedModel
 from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
 from dataquality.schemas.split import Split
 from dataquality.schemas.task_type import TaskType
-from dataquality.utils.helpers import get_task_type
+from dataquality.utils.task_helpers import get_task_type
 
 
 def watch(
--- a/dataquality/loggers/data_logger/base_data_logger.py
+++ b/dataquality/loggers/data_logger/base_data_logger.py
@@ -317,6 +317,9 @@ class BaseGalileoDataLogger(BaseGalileoLogger):
             return
         in_frame_split = vaex.open(f"{in_frame_path}/*.{self.INPUT_DATA_FILE_EXT}")
         in_frame_split = self.convert_large_string(in_frame_split)
+        # Already has 'text' column (which is the input data)
+        # TODO: Iterate through all samples in in_frame_split
+        # and run the text through the tokenizer and model
         self.upload_split_from_in_frame(
             object_store,
             in_frame_split,
--- a/dataquality/loggers/data_logger/base_data_logger.py
+++ b/dataquality/loggers/data_logger/base_data_logger.py
@@ -317,9 +317,6 @@ class BaseGalileoDataLogger(BaseGalileoLogger):
             return
         in_frame_split = vaex.open(f"{in_frame_path}/*.{self.INPUT_DATA_FILE_EXT}")
         in_frame_split = self.convert_large_string(in_frame_split)
-        # Already has 'text' column (which is the input data)
-        # TODO: Iterate through all samples in in_frame_split
-        # and run the text through the tokenizer and model
         self.upload_split_from_in_frame(
             object_store,
             in_frame_split,
--- a/dataquality/loggers/data_logger/seq2seq.py
+++ b/dataquality/loggers/data_logger/seq2seq.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Any, List, Optional, Set, Tuple, Union, cast
 
 import pandas as pd
 import pyarrow as pa
+import torch
 import vaex
 from vaex.dataframe import DataFrame
 
@@ -12,12 +13,16 @@ from dataquality.loggers.data_logger.base_data_logger import (
     DataSet,
     MetasType,
 )
-from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
+from dataquality.loggers.logger_config.seq2seq import (
+    Seq2SeqLoggerConfig,
+    seq2seq_logger_config,
+)
 from dataquality.schemas.dataframe import BaseLoggerDataFrames
 from dataquality.schemas.seq2seq import Seq2SeqInputCols as C
 from dataquality.schemas.split import Split
-from dataquality.utils.seq2seq import (
+from dataquality.utils.seq2seq import (  # noqa needed to register the vaex function
     align_tokens_to_character_spans,
+    generate_output,  # noqa: F401
 )
 from dataquality.utils.vaex import rename_df
 
@@ -67,7 +72,7 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
     """
 
     __logger_name__ = "seq2seq"
-    logger_config = seq2seq_logger_config
+    logger_config: Seq2SeqLoggerConfig = seq2seq_logger_config
     DATA_FOLDER_EXTENSION = {"emb": "hdf5", "prob": "hdf5", "data": "arrow"}
 
     def __init__(self, meta: Optional[MetasType] = None) -> None:
@@ -193,6 +198,68 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
     @classmethod
     def _get_prob_cols(cls) -> List[str]:
         return ["id"]
+
+    @classmethod
+    def create_in_out_frames(
+        cls,
+        in_frame: DataFrame,
+        dir_name: str,
+        prob_only: bool,
+        split: str,
+        epoch_or_inf: Union[str, int],
+    ) -> BaseLoggerDataFrames:
+        """Formats the input data and model output data
+
+        In this step, we concatenate the many hdf5 files created during model training
+        and logging. We log those in threaded processes, and here we combine them
+        into a single hdf5 file that vaex can read into a dataframe
+
+        :param in_frame: the input dataframe
+        :param dir_name: The directory of all of the output hdf5 files
+        :param prob_only: If we are only uploading probability data. We only upload
+            probability data for all epochs except the last one (we dont use cross-epoch
+            embeddings currently, so we dont log them)
+        :param split: The split we are logging for
+        :param epoch_or_inf: The epoch or inference name we are logging for
+        """
+        in_frame = cls.add_generated_output_to_df(in_frame, split)
+        return super().create_in_out_frames(
+            in_frame, dir_name, prob_only, split, epoch_or_inf
+        )
+
+    @classmethod
+    def add_generated_output_to_df(cls, df: DataFrame, split: str) -> DataFrame:
+        """Adds the generated output to the dataframe
+
+        Adds the generated output to the dataframe, and also adds the
+        `token_label_positions` column
+        """
+        logger_config = cls.logger_config
+        tokenizer = logger_config.tokenizer
+        model = logger_config.model
+        generation_config = logger_config.generation_config
+        if tokenizer is None:
+            raise GalileoException(
+                "You must set your tokenizer before logging. Use `dq.set_tokenizer`"
+            )
+        if model is None:
+            raise GalileoException(
+                "You must set your model before logging. Use `watch`"
+            )
+        if generation_config is None:
+            raise GalileoException(
+                "You must set your generation config before logging. Use `watch`"
+            )
+        if split not in logger_config.generation_splits:
+            print("Skipping generation for split", split)
+            return df
+
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        model.eval()
+        df[C.generated_output.value] = df[C.text.value].func.generate_output(
+            tokenizer, model, device, generation_config
+        )
+        return df
 
     @classmethod
     def separate_dataframe(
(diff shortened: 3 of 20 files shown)


elboy3:
Not even close to being ready to be reviewed

dataquality/utils/helpers.py
```diff
@@ -133,3 +135,13 @@ def mps_available() -> bool:
...
+
+
+def get_task_type(task_type: Optional[TaskType] = None) -> TaskType:
+    task = task_type or config.task_type
+    if not task:
+        raise GalileoException(
+            "You must provide either a task_type or first call "
+            "dataqualtiy.init and provide one"
+        )
+    return task
```
elboy3, line 147: moved from `dataquality/core/log.py`

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -194,6 +199,68 @@ def get_valid_attributes() -> List[str]:
...
+        if generation_config is None:
+            raise GalileoException(
+                "You must set your generation config before logging. Use `watch`"
+            )
+        if split not in logger_config.generation_splits:
+            print("Skipping generation for split", split)
+            return df
+
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        model.eval()
```
elboy3, line 258: torch.no_grad

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -194,6 +199,68 @@ def get_valid_attributes() -> List[str]:
...
+        and logging. We log those in threaded processes, and here we combine them
+        into a single hdf5 file that vaex can read into a dataframe
+
+        :param in_frame: the input dataframe
+        :param dir_name: The directory of all of the output hdf5 files
+        :param prob_only: If we are only uploading probability data. We only upload
+            probability data for all epochs except the last one (we dont use cross-epoch
+            embeddings currently, so we dont log them)
+        :param split: The split we are logging for
+        :param epoch_or_inf: The epoch or inference name we are logging for
```
elboy3, line 223: update :)

dataquality/loggers/data_logger/base_data_logger.py
```diff
@@ -317,6 +317,9 @@ def upload_split(
             return
         in_frame_split = vaex.open(f"{in_frame_path}/*.{self.INPUT_DATA_FILE_EXT}")
         in_frame_split = self.convert_large_string(in_frame_split)
+        # Already has 'text' column (which is the input data)
+        # TODO: Iterate through all samples in in_frame_split
+        # and run the text through the tokenizer and model
```
elboy3, line 322: remove

Ben-Epstein (changes requested):
tests??!?!?!?

dataquality/utils/seq2seq.py
```diff
@@ -3,9 +3,55 @@
...
+            # Shape - [1, seq_len]
+            input_ids = tokenizer(str(sample), return_tensors="pt")["input_ids"].to(
+                device
+            )
+            generation_respone = model.to(device).generate(
+                input_ids=input_ids,
+                generation_config=generation_config,
+                return_dict_in_generate=True,  # Required to get token probs
+                output_scores=True,  # Required to get token probs
+            )
```
setu4993, line 35: I'd honestly expect this to be a fairly expensive (time + compute) process that scales linearly with the dataset.

Also, if we're not hooking up with the user's own generate fn, we'll have a different output than what maybe they're looking at (temperature, num beams, etc.).

dataquality/utils/seq2seq.py
```diff
@@ -3,9 +3,55 @@
...
+
+            # Remove the <pad> token to seed generation
+            generated_tokens = generation_respone.sequences[0, 1:]
```
setu4993, line 38: Why? I think we'd want it because that's what tells the model to start generating...

dataquality/utils/seq2seq.py
```diff
@@ -3,9 +3,55 @@
...
+            # TODO: Skip logits for now, will update later
+            # generated_logits = torch.stack(generation_respone.scores)[:, 0, :]
+
+            # Note that the model may also end with the <eos> token. We should
+            # check for this to get the correct logits!
+            if generated_tokens[-1] == tokenizer.eos_token_id:
+                # Cut off the generated <eos> token if need
+                # And remove the probability associated with its generation
+                generated_tokens = generated_tokens[:-1]
+                # generated_logits = generation_respone.scores[:-1]
```
setu4993, line 48: Curious about how this impacts our calculations? Do we also drop it when calculating DEP?

dataquality/utils/seq2seq.py
```diff
@@ -3,9 +3,55 @@
...
+
+def add_generated_output_to_df(
+    df: vaex.DataFrame,
+    tokenizer: PreTrainedTokenizerFast,
+    model: PreTrainedModel,
+    device: torch.device,
+    generation_config: GenerationConfig,
+) -> vaex.DataFrame:
+    @vaex.register_function()
+    def generate_output(text: pa.array) -> np.ndarray:
```
setu4993, line 23: How is text `pa.array` 🤔? Maybe this warrants naming it as something different since it's tokenized input?

dataquality/schemas/seq2seq.py
```diff
@@ -26,15 +26,15 @@ def set_cols(cls, df: DataFrame) -> DataFrame:
     def set_input(cls, df: DataFrame) -> DataFrame:
         """Sets the input column for the dataframe"""
         if cls.text.value in df.get_column_names():
-            df = df.rename(cls.text.value, cls.input.value)
+            df.rename(cls.text.value, cls.input.value)
```
setu4993, line 29: Neat (and fallible) that `vaex` renames inplace.

dataquality/integrations/seq2seq/hf.py
```diff
@@ -0,0 +1,33 @@
...
+    """Seq2seq only. Log model generations for your run
+
+    Iterates over a given dataset and logs the generations for each sample.
+    `model` must be an instance of transformers PreTrainedModel and have a `generate`
+      method.
+    """
+    task_type = get_task_type()
+    assert task_type == TaskType.seq2seq, "This method is only supported for seq2seq"
+    assert isinstance(
+        model, PreTrainedModel
```
setu4993, line 23: Do we need to limit to models that can generate?

dataquality/integrations/seq2seq/hf.py
```diff
@@ -0,0 +1,33 @@
...
+    ), "model must be an instance of transformers PreTrainedModel"
+    assert model.can_generate(), "model must contain a `generate` method for seq2seq"
```
setu4993, line 25: Ah, never mind :).

setu4993 (commented):
Left a few questions below, let's chat more tomorrow.

dataquality/integrations/seq2seq/hf.py
```diff
@@ -0,0 +1,33 @@
...
+
+    seq2seq_logger_config.model = model
+    seq2seq_logger_config.generation_config = generation_config
+
+    generation_splits = {Split.validation, Split.test}
+    if generate_training_data:
```
jonathangomesselman, line 31: We should maybe put a warning / flag here that clearly notifies the user `Hey you're choosing to generate on the Training Data which may take quite a bit of time!`

dataquality/utils/seq2seq.py
jonathangomesselman: I think I agree there. In other situations we have added `tokenized` to the name?

jonathangomesselman: @setu4993 I think it is a `pa.array[string]` - i.e. a list of text strings. Maybe this would be better called `texts`

dataquality/utils/seq2seq.py
jonathangomesselman: Yes this will definitely be an expensive process. I think that is why we should especially warn users when they indicate that they want to generate over the training data.

jonathangomesselman: Super valid point about the generate function. Here are some of my thoughts on it based on some research into how different repos use hf's generate + how I have been using it.

Essentially, it seems like a pretty common generation pattern is to just rely on hf's `generate` function, which is built in to all generative models. Pretty classically (and great for us), hf has made the generation process very standardized and simple, abstracting much of the work away from the user. The key lies in the `generation_config`, which allows you to specify parameters such as `temperature`, `top_k`, `top_p`, ... even things like `num_beams`.

The `num_beams` is a great point though. My thinking was that for our first pass at this we should skip support for beam based generation. We can certainly handle this down the road, it just adds a small bit of complexity in how the `generation_response` is formatted. I think generally we would just have to have a switch case to process `generation_response` differently based on its returned `type` (which changes based on if you use `beam_search`).

How does this sound @setu4993? There is definitely then the possibility that they are using something completely different, which hf does not support. My hope is this will not happen often and if it does we can cross that bridge when it does?

dataquality/utils/seq2seq.py
jonathangomesselman: We definitely want it when we pass the `input_ids` to the model, but when we process the generated response I think we would want to remove it. As a user I wouldn't expect to see the `pad` token - as it is somewhat implied as the start of the generation.

dataquality/utils/seq2seq.py
jonathangomesselman: My thinking was yes! This was something I was not 100% sure about, but generally I have seen people remove special tokens when they for example visualize generated output. 

In general, my thinking was that we would not show probs / text for the <eos> token. However, it can be something valuable, especially if you are fine-tuning your model with the expectation that it will stop generating in specific situations and this is not happening. So visualizing the model's log probs over the <eos> token could be interesting - though I have not seen this really done.

dataquality/loggers/data_logger/seq2seq.py
jonathangomesselman: Yes! You can also add this fun little decorator `@torch.no_grad()`. I have seen it used extensively in the hf repo to avoid endlessly having to do `with torch.no_gra():`

jonathangomesselman:
@setu4993 - Added responses to your questions. Happy to chat more through some of the design choices. Thanks for the help here!

dataquality/utils/seq2seq.py
setu4993: Generation is expensive regardless of the data split it is running on, which is why we shouldn't regenerate (as we're doing here) but use what the user generates as a part of their training / validation / evaluation loop.

> Essentially, it seems like a pretty common generation pattern is to just rely on hf's `generate` function, which is built in to all generative models. Pretty classically (and great for us), hf has made the generation process very standardized and simple, abstracting much of the work away from the user. The key lies in the `generation_config`, which allows you to specify parameters such as `temperature`, `top_k`, `top_p`, ... even things like `num_beams`.

Yeah, I know :). I've been using it for ~4 years and it's pretty neat to have a single API to use!

> The `num_beams` is a great point though. My thinking was that for our first pass at this we should skip support for beam based generation. We can certainly handle this down the road, it just adds a small bit of complexity in how the `generation_response` is formatted. I think generally we would just have to have a switch case to process `generation_response` differently based on its returned `type` (which changes based on if you use `beam_search`).

I don't really follow how the type changes if we use beam search 🤔. From what I remember and ([just confirmed from the docs](https://huggingface.co/docs/transformers/v4.31.0/en/generation_strategies#beamsearch-decoding)) the dtype remains the same: `str` regardless of beam search (`list` is because it's a "batch" being decoded), since beam search is an internal decoding detail and not surfaced to the user as multiple responses.

> How does this sound @setu4993? There is definitely then the possibility that they are using something completely different, which hf does not support. My hope is this will not happen often and if it does we can cross that bridge when it does?

My biggest concern is us _duplicating_ the generation step, which I am certain we shouldn't do. This is where we get mismatches between what the user is doing vs. what we are doing and can lead users to lose trust in the product because, well, we aren't showing them the outputs they see.

If this were text classification, this would be equivalent to us showing different classes than the actual prediction. That's the impact of changing _anything_ about the generation strategy.

dataquality/utils/seq2seq.py
setu4993: Yea, but `tokenizer.decode` would simply take it out :).

We shouldn't need to do special handling.

dataquality/utils/seq2seq.py
setu4993: > generally I have seen people remove special tokens when they for example visualize generated output.

Yes, but it doesn't need to be handled separately outside using the defaults (see ^ comment).

If we want to drop something, let's use a mask instead of relying on slicing.

dataquality/schemas/seq2seq.py
setu4993: @elboy3 : `pandas`, `pyarrow`, etc. do not perform operations like rename on an existing DF but on a copy. That prevents inadvertent issues from occurring.

When I originally read the removed line, what I got was: We were following the `pandas` pattern and creating a new DF with renamed columns. It being modified here led me to the docs, which helped me realize the OP :).

dataquality/utils/seq2seq.py
elboy3: @setu4993 so the main requested change here is that instead of doing a final pass over the data to log the model's post training "generated output" for the split, we should log the generated output at each epoch and just use the output for the last epoch as generated ?


what outputs are different btwn what we show them and what they see (what do you mean by "outputs they" see by the way?)

dataquality/utils/seq2seq.py
elboy3: Yup, I think the typing should actually be `pa.StringArray`, will fix

dataquality/utils/seq2seq.py
setu4993: The main requested change is that instead of us _generating_ the output, we should hook into the user's process for generating those outputs.

By "outputs they see" I mean the outputs the user generates as a part of their validation loop.

dataquality/utils/seq2seq.py
jonathangomesselman: Good point. @elboy3 for the decoding step in the tokenizer we can avoid this by doing `tokenizer.decoder(..., skip_special_tokens=True)`

dataquality/utils/seq2seq.py
jonathangomesselman: For the decoding this is a great point! 

But when we also are logging / handling the `logits` I think that we will want to strip the `<eos>` token probability. Unless we think that this is actually something we want to surface to the user. Which, I kind of go back and forth on. If their tokenizer is adding it than I may almost be learning more on the side of having us show it. But I think that this could be a bigger discussion that I will post in the ml channel.

jonathangomesselman: > If we want to drop something, let's use a mask instead of relying on slicing.

For this do you mean using something like `generated_tokens = generated_tokens[np.where(generated_tokens != eos_token_id)[0]]`? Is the idea that a mask is a bit more precise than slicing?

dataquality/utils/seq2seq.py
jonathangomesselman: > I don't really follow how the type changes if we use beam search 🤔. From what I remember and ([just confirmed from the docs](https://huggingface.co/docs/transformers/v4.31.0/en/generation_strategies#beamsearch-decoding)) the dtype remains the same: str regardless of beam search (list is because it's a "batch" being decoded), since beam search is an internal decoding detail and not surfaced to the user as multiple responses.

The difference comes from when you want to get back the model logits - i.e. when you pass in `return_dict_in_generate=True` and `output_scores=True`. In this situation the beams are inadvertently surfaced to the user as multiple probabilities vectors (beams) per token. Looking at the [docs here](https://huggingface.co/docs/transformers/v4.30.0/en/main_classes/text_generation#transformers.GenerationMixin.generate) you see that we will get a `Beam[Search/Sampler][EncoderDecoder/DecoderOnly]Output`. While this is certainly something we can handle, it does add complexity on the DQ side for computing probability based metrics, which is why I suggested we maybe add it in V1.

jonathangomesselman: @setu4993 One thing for example with fine-tuning is I don't think that users are always generating at each step in their validation loop. Since in general metrics for generative tasks can be quite weak, I think often people just look at perplexity as a proxy for val performance. However, the flip side is certainly that val perplexity does not always correlate with human rater eval (e.g. discussed in [Lima](https://arxiv.org/pdf/2305.11206.pdf) where they use manual selection of a checkpoint based on a tiny held out set). 

Maybe we can have a bigger discussion about this not over messages. I am not sure I see why their would be mismatch + I think we are ignoring some of the benefits to abstracting away the generation step. In general users face a lot of friction integrating Galileo, and with how standard generation is for LLMs in hf, we get a lot of wins when doing this for the user. As an example, many of the eval libraries take a model and handle the generation for the user.

setu4993: > I am not sure I see why their would be mismatch

Here's a typical LLM finetuning workflow that I've performed for years + seen others do a bunch:
1. Train a model.
2. Perform generation during the validation step with default generation config.
3. Generate during the last validation step with custom generation config.
4. Generate during the test step with custom generation config.

What we're doing here is using the default generation config at step 3, which IMHO would cause a mismatch because we're using the default / Galileo specific config rather than one that the user would use to evaluate their own model.

When I'm generating my final set of generated outputs, you're right, I may not use the final model and a previous checkpoint, which is even more so why I don't think we should generate for the user independent of the generation steps they're using as a part of their existing training scripts.

The generation is opinionated, because it has various configurations that completely change the actual result of the generation, which causes a mismatch between what I'd get if I ran the generation myself on a trained model vs. what I'd see on Galileo.

A good middle ground could be to simply let the users log their generation instead of doing the generation _for them_.

Happy to chat about it sync anytime if it'd help!

jonathangomesselman: I would agree with that work flow! I think there may be some tradeoffs in whether people do full generation during fine-tuning, but I think that is a small detail.

> What we're doing here is using the default generation config at step 3, which IMHO would cause a mismatch because we're using the default / Galileo specific config rather than one that the user would use to evaluate their own model.

I think that this is maybe where we are just slightly mis-communicating. I am certainly not suggesting that we use a `default Galileo generation config`. This would be something that the user provides to us, to as you say, fully control the final generation step over their `validation` and `test` data. 

I think that a fair question is around generating based on different checkpoints. But again I think there is something to be said about ease of utilizing Galileo + working towards an first working version of DQ. In terms of the first (ease of use), if the user can finish training and super easily be able to log generated output to Galileo that seems like a pretty big win - again we obviously are sacrificing some customizability, but as we have seen tons of users have loved `dq.auto`. On the second (building towards a first working version), this is also certainly the easier first assumption, since otherwise there is much more nuance around getting the users `logits`, etc. 

I definitely think that at some point we should support user logged generations! I just don't know if I agree that it is an early middle ground? It just seems like the other end of the spectrum.

dataquality/utils/seq2seq.py
jonathangomesselman: My only thinking here is that if these tokens appear at other indices, it may signal something unexpected. In this case, we may even want to throw some flag or something.

elboy3:
We have merged this work in separate PRs, closing!

PR #737, opened 2023-08-11
Added jsl evaluate

franz101:
First draft

### DIFF
--- a/dataquality/integrations/jsl.py
+++ b/dataquality/integrations/jsl.py
@@ -1,0 +1,120 @@
+from typing import Optional
+from pyspark.sql import DataFrame
+
+from pyspark.sql import Window
+from pyspark.sql.functions import col, monotonically_increasing_id, row_number, size
+
+import dataquality as dq
+from dataquality.schemas.task_type import TaskType
+from dataquality.utils.jsl import (
+    PipelineLike,
+    add_labels_df,
+    align_gold_spans,
+    find_ner_model,
+    get_relevant_cols,
+    get_stages,
+    pad_df,
+    prepare_df_for_dq,
+)
+
+chunksize = 512
+max_token_length = 512
+
+
+class Project:
+    def __init__(
+        self,
+        project_name: str,
+        run_name: str,
+        task_type: TaskType = TaskType.text_ner,
+        url: Optional[str] = None,
+    ) -> None:
+        self.project_name = project_name
+        self.run_name = run_name
+        self.task_type = task_type
+        self.url = url
+        self.setup()
+
+    def setup(self) -> None:
+        if self.url:
+            dq.set_console_url(self.url)
+        dq.init(
+            project_name=self.project_name,
+            run_name=self.run_name,
+            task_type=self.task_type,
+        )
+
+    def watch(self, pipeline: PipelineLike, text_col: str = "text") -> None:
+        model = find_ner_model(pipeline, text_col)
+        model.setIncludeAllConfidenceScores(True)
+        model.setIncludeConfidence(True)
+        self.emb_col, _, _ = get_relevant_cols()
+        self.text_col = text_col
+        self.pipeline = pipeline
+
+    def evaluate(
+        self,
+        df: DataFrame,
+        split: str = "training",
+        finish: bool = False,
+        pipeline: Optional[PipelineLike] = None,
+        batch_size: int = 2048,
+        max_token_length: int = 1024,
+    ) -> Optional[str]:
+        if pipeline:
+            self.watch(pipeline)
+        if not self.pipeline:
+            raise ValueError("No pipeline found. Please call watch(pipeline) first.")
+        df = df.withColumn(
+            "id", row_number().over(Window.orderBy(monotonically_increasing_id())) - 1
+        )
+
+        df = df.withColumn("token_length", size("token"))
+        ds_len = df.count()
+        split = "training"
+        chunksize = batch_size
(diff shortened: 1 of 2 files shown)


dataquality/integrations/jsl.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        self.project_name = project_name
+        self.run_name = run_name
+        self.task_type = task_type
+        self.url = url
+        self.setup()
+        if pipeline:
+            self.watch(pipeline)
+        if dataset and pipeline:
+            self.evaluate(dataset, finish=finish)
```
bogdan-galileo, line 63:
```suggestion
        if pipeline is not None:
            self.watch(pipeline)
            if dataset is not None:
                self.evaluate(dataset, finish=finish)
```
safer ?

dataquality/integrations/jsl.py
```diff
@@ -0,0 +1,170 @@
...
+
+        df = df.withColumn("token_length", size("token"))
+        ds_len = df.count()
+        split = Split.training
+        chunksize = batch_size
+        for chunk in range(0, ds_len, chunksize):
+            df_chunk = df.filter(
+                (col("id") >= chunk)
+                & (col("id") < (chunk + chunksize))
+                & (col("token_length") < max_token_length)
```
bogdan-galileo, line 126: If I understand correctly we don't even take the rows that have too many tokens ? You wouldn't prefer to truncate ?

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+
+def extract_ner_result(ner_list: List[Row]) -> List:
+    """Extracts the NER result from a list of NER results
+    :param ner_list: A list of NER results
+    :return: A list of strings, where each string is the NER result of a token
+    """
+    return [row.result for row in ner_list]
+
+
+def extract_ner_gold_spans(ner_list: List[Row]) -> List[dict]:
```
bogdan-galileo, line 55: Don't we alraedy have that function somewhere ? I remember it was my coding interview with Nidhi and I think she mentioned this already exists somewhere!

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+    """Extracts the NER gold spans from a list of NER results
+    :param ner_list: A list of NER results
+    :return: A list of dictionaries, where each dictionary contains the start, end
+        and label of a gold span
+    """
+    spans = []
+    current_span = None
+    for row in ner_list:
+        # Check if the row is a beginning of a named entity (B-)
+        if "B-" in row.result:
```
bogdan-galileo, line 65: should we check that it starts with B- and not that it contains it ?

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+            # If there is an existing span, we add it to the list
+            if current_span:
+                spans.append(current_span)
+            # Start a new span
+            current_span = {
+                "start": row.begin,
+                "end": row.end + 1,
+                "label": row.result.split("-")[-1],
```
bogdan-galileo, line 73: Should we be careful if the rest of the span contains `-` ?

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+    df: pd.DataFrame, emb_col: str = "embeddings", label_col: str = "ner_list"
+) -> pd.DataFrame:
+    """Pads a dataframe
+    :param df: The pandas dataframe to be padded
+    :param emb_col: The name of the embeddings column (default: embeddings)
+    :param label_col: The name of the labels column (default: ner_list)
+    :return: The padded dataframe
+    """
+    global MAX_TOKEN_LEN
+    max_length = MAX_TOKEN_LEN
```
bogdan-galileo, line 175: maybe better to have as an arg with default value MAX_TOKEN_LEN

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+            # If there's an active span, we add it to the list and reset current_span
+            if current_span:
+                spans.append(current_span)
+                current_span = None
+
+    # Add any remaining span after the loop
+    if current_span:
+        spans.append(current_span)
+
+    return spans
```
bogdan-galileo, line 92: Tests would help for a few of these complicated functions to make sure they work as expected

dataquality/utils/jsl.py
```diff
@@ -0,0 +1,406 @@
...
+            mode="constant",
+            constant_values=0,
+        )
+        for row in df[label_col]
+    ]
+    return df
+
+
+# Function to align gold spans with token boundaries
+def align_gold_spans(
```
bogdan-galileo, line 198: test would help here too

bogdan-galileo (approved):

bogdan-galileo (approved):
Tests would help for a few of these complicated token-level functions

dataquality/integrations/jsl.py
franz101: it's pretty much the same as an object except empty list or dictionary == False.
And we always watch first before we evaluate.

dataquality/utils/jsl.py
franz101: Oh, maybe, this in PySpark though. Curios if you can point me to the code.

dataquality/utils/jsl.py
franz101: good point. fixed it.

dcaustin33 (approved):

PR #739, opened 2023-08-23
fix: Remove unneeded column in s2s df

elboy3:
As per feedback on https://github.com/rungalileo/rungalileo/pull/792/files, we:
- rename `target_output` -> `target`
- no longer log `tokenized_label` in finalized df

Same thing happening here: https://github.com/rungalileo/runners/pull/423


setu4993 (approved):
Looks good. I'll let you figure out what's up with tests...

elboy3:
> Looks good. I'll let you figure out what's up with tests...

linting issue, fixed

PR #742, opened 2023-09-13
feat: refactor logprobs

jonathangomesselman:

This PR re-factors the model logging code for Seq2Seq and adds support for `logprobs` and `top_logprobs`. With this refactor we move several helper functions to `utils/seq2seq.py` to set ourselves up to more seamlessly share code when implementing logging for model generation

Key additions that we add to the model logging schema are:
- `perplexity`: per sample perplexity; though this will quickly be deprecated in favor of computing in runners.
- `token_logprobs`: per-token `logprobs` for the gold label.
- `top_logprobs`: the top K=5 (predicted tokens (str), logprobs (float)) for each token

Note that we have fully moved away from `token_probs` in favor of storing and working with `logprobs`.

For `top_logprobs`, we specify use the following format for a _single_ sample
```
top_logprobs: List[List[Tuple[str, float]]] = [
        [("the", -0.2), ("a", -0.6), ...],
        [("cat", -0.05), ("dog", -0.1), ...],
        ...
    ]
```
We handle this custom pyarrow format as
```
TOP_LOGPROBS_SCHEMA = pa.list_(pa.map_(pa.string(), pa.float32()))
top_logprobs = pa.array(top_logprobs, type=TOP_LOGPROBS_SCHEMA)
```

Last thing to mentions is that we have left in the code for computing `DEP` and `Perplexity` - though these will be removed shortly as that logic is moved to runners.


dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -39,9 +42,12 @@ def __init__(
             inference_name=inference_name,
             labels=labels,
         )
+        # TODO Move DEP and Perplexity computation outside of DQ
```
elboy3, line 45: 🧹

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -39,9 +42,12 @@ def __init__(
...
         self.sample_dep: List[float] = []
-        self.token_dep = pa.array([])
-        self.token_gold_probs = pa.array([])
+        self.sample_perplexity: List[float] = []
+        self.token_deps = pa.array([])
```
elboy3, line 48: so let's remove this now, right? since bogdans work just got merged

dataquality/schemas/seq2seq.py
```diff
@@ -41,9 +47,11 @@ def set_target(cls, df: DataFrame) -> DataFrame:
 
 class Seq2SeqOutputCols(str, Enum):
     id = "id"
-    dep = "data_error_potential"
-    token_deps = "token_deps"
-    token_gold_probs = "token_gold_probs"
+    dep = "data_error_potential"  # TODO Depricate
+    perplexity = "perplexity"  # TODO Depricate
+    token_deps = "token_deps"  # TODO Depricate
```
elboy3, line 52:
```suggestion
    perplexity = "perplexity"  # TODO Depricate
```

nice. good to deprecate the dep ones

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,152 @@
...
+
+    Post-conditions:
+    ----------------
+        - logprobs is left unchanged
+
+    TODO this can be so so much faster with torch on gpu!
+
+    Parameters:
+    -----------
+    logprobs: np.ndarray of shape [(optional)batch_size, seq_len, vocab_size]
```
elboy3, line 31: if its optional will batch size be 1 or it would just be 2d instead of 3d?

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,152 @@
...
+        all but the last dimension (i.e. not -1) to seemlisly handle
+        samples or batches.
+    """
+    # Multiply by -1 to reverse the order
+    logprobs *= -1
+    partitioned_logprob_indices = np.argpartition(logprobs, TOP_K, axis=-1)
+    top_logprob_indices = partitioned_logprob_indices[..., :TOP_K]
+    logprobs *= -1
+
+    return top_logprob_indices
```
elboy3, line 47: 🔥

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +247,53 @@ def _get_data_dict(self) -> Dict:
...
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Remove padding tokens from logprobs and top_indices
+
+        To remove padding we use the tokenized labels and slice
+        tokens depending on the padding side of the tokenizer.
+
+        Parameters:
+        -----------
+
+        """
```
elboy3, line 272: TODO?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +247,53 @@ def _get_data_dict(self) -> Dict:
...
+            self.logger_config.id_to_tokens[self.token_map_key][sample_id]
+        )
+        return labels
+
+    def _remove_padding(
+        self,
+        labels: np.ndarray,
+        sample_logprobs: np.ndarray,
+        sample_top_indices: np.ndarray,
+    ) -> Tuple[np.ndarray, np.ndarray]:
```
elboy3, line 263: so this is just for a single sample? can u clarify in docstring or in fn name?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -63,103 +69,175 @@ def validate_and_format(self) -> None:
...
+        )
 
     def _get_data_dict(self) -> Dict:
         """Returns the data dictionary for writing to disk"""
         batch_size = len(self.ids)
         data = {
             C.id.value: self.ids,
-            C.token_deps.value: self.token_dep,
+            C.token_deps.value: self.token_deps,
             C.dep.value: self.sample_dep,
```
elboy3, line 237: remove! 🧹

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,152 @@
...
+
+    # Extract token_logprobs - shape [len(labels)]
+    label_logprobs = np.take_along_axis(
+        sample_logprobs, sample_labels, axis=-1
+    ).squeeze()
+
+    # Compute top_logprobs
+    top_logprobs = extract_top_logprobs(sample_logprobs, sample_top_indices, tokenizer)
+
+    return label_logprobs, top_logprobs
```
elboy3, line 152: note we're calling them `token_logprobs` everywhere else, should probably here too for consistency

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,152 @@
...
+    Returns:
+    --------
+    label_logprobs: np.ndarray of shape [seq_len]
+        Label logprobs for each token in the sample
+    top_logprobs: List[Dict[str, float]]
+        List of top-k (str) predictions + corresponding logprobs
+    """
+    # Ensure final shape - [len(labels), 1]
+    if sample_labels.ndim != 1:
+        raise GalileoException("Expects sample_labels to be a 1D array")
```
elboy3, line 141:
```suggestion
    if sample_labels.ndim != 1:
        raise GalileoException(f"Invalid shape {sample_lables.shape}, process_sample_logprobs expects sample_labels to be a 1D array")
```

you'll have to format ^

elboy3 (commented):
truck ya! took a quick glance, will take a deeper look later today. very profesh PR looking sick

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: Clarifying in DocString

dataquality/schemas/seq2seq.py
jonathangomesselman: Wait keep perplexity for now?

jonathangomesselman: I guess I can keep for now. But we should in the future move to runners. I will create a quick ticket.

jonathangomesselman: https://app.shortcut.com/galileo/story/7593

dataquality/utils/seq2seq.py
jonathangomesselman: It would be 2D instead of 3D. During `model_logging` the shape is `[bs, seq_len, vc]` vs. during `generation` it is `[seq_len, vc]`

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,155 @@
...
+    In both situations, we compute the top logprobs along the final (-1)
+    vocab dimension. We use `np.argpartition` to remove the overhead of
+    sorting along the vocab dimension - O(nlog(n)) -> O(n).
+    For reference see: https://numpy.org/doc/stable/reference/generated/numpy.argpartition.html
+
+    Post-conditions:
+    ----------------
+        - logprobs is left unchanged
+
+    TODO this can be so so much faster with torch on gpu!
```
elboy3, line 27: let's make a shortcut ticket to track this! i think we should definitely leverage this speed up (eventually)

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,155 @@
...
+
+    return top_logprobs
+
+
+def process_sample_logprobs(
+    sample_logprobs: np.ndarray,
+    sample_labels: np.ndarray,
+    sample_top_indices: np.ndarray,
+    tokenizer: PreTrainedTokenizerFast,
+) -> Tuple[np.ndarray, List[List[Tuple[str, float]]]]:
```
elboy3, line 108: note to self - i wonder if we should use a pydantic class to represent the return and the log probs

dataquality/utils/seq2seq.py
```diff
@@ -4,8 +4,155 @@
...
+    sample_top_logprobs = np.take_along_axis(sample_logprobs, top_indices, axis=-1)
+
+    # Generate top_k (string, logprob) token by token
+    top_logprobs: List[List[Tuple[str, float]]] = []
+    for token_top_ids, token_top_logprobs in zip(top_indices, sample_top_logprobs):
+        # List of Tuple[str, int] --> (token string, logprob)
+        token_top_logprobs_mapping = []
+        # Loop over the top_k predictions for the given token position
+        for pred_token_id, logprob in zip(token_top_ids, token_top_logprobs):
+            str_token = tokenizer.decode(pred_token_id)
```
elboy3, line 95: how fast is decode? do you think we'd get any valuable speed ups from calling decode on the full list of token ids?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +169,74 @@ def _get_data_dict(self) -> Dict:
...
+        -------
+        sliced_sequences: Tuple[np.ndarray, ...] - each array has shape [n_tokens, ...]
+            Returns a tuple with the padding tokens removed for each
+            token sequence in *args - maintaining order and non-token dimensions.
+        """
+        # Remove padding based on the padding_side of the tokenizer
+        num_tokens = len(labels)
+        sliced_sequences: Tuple[np.ndarray, ...] = ()
+        for token_sequence in args:
+            if self.logger_config.tokenizer.padding_side == "left":  # type: ignore
```
elboy3, line 220: we should pull this out of the for loop since it will never change 

we could even pull it out of `_remove_padding` and pass it into the fn as an arg, your call

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +169,74 @@ def _get_data_dict(self) -> Dict:
...
+
+        Returns:
+        -------
+        sliced_sequences: Tuple[np.ndarray, ...] - each array has shape [n_tokens, ...]
+            Returns a tuple with the padding tokens removed for each
+            token sequence in *args - maintaining order and non-token dimensions.
+        """
+        # Remove padding based on the padding_side of the tokenizer
+        num_tokens = len(labels)
+        sliced_sequences: Tuple[np.ndarray, ...] = ()
```
elboy3, line 218: let's discuss Tuple vs List here, List might make more sense since tuples are immutable

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +169,74 @@ def _get_data_dict(self) -> Dict:
 
     def _write_dict_to_disk(self, path: str, object_name: str, data: Dict) -> None:
         save_arrow_file(path, object_name, data)
+
+    def _retrieve_sample_labels(self, sample_id: int) -> np.ndarray:
+        """Retrieve the labels array based on the sample it"""
```
elboy3, line 174: sample id?

elboy3: i know it's annoying but maybe give a small example in the docstring?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -169,3 +169,74 @@ def _get_data_dict(self) -> Dict:
...
+            self.logger_config.id_to_tokens[self.token_map_key][sample_id]
+        )
+        return labels
+
+    def _remove_padding(
+        self,
+        labels: np.ndarray,
+        *args: np.ndarray
+        # sample_logprobs: np.ndarray,
+        # sample_top_indices: np.ndarray,
```
elboy3, line 185: 🧹 ?

dataquality/utils/seq2seq.py
jonathangomesselman: I think the top_logprobs would definitely make sense. This type is ugly!! 

I can take a stab at this!

jonathangomesselman: Less certain about the return?

jonathangomesselman: Gave an attempt at this! Not sure if this is exactly the way to do it though.

dataquality/utils/seq2seq.py
bogdan-galileo:
```
str_tokens = tokenizer.decode(token_top_ids)
token_top_logprobs_mapping = list(zip(str_tokens, token_top_logprobs))
```
?

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -63,103 +66,100 @@ def validate_and_format(self) -> None:
...
+                sample_top_indices=sample_top_indices,
+                tokenizer=self.logger_config.tokenizer,
+            )
+
+            batch_token_logprobs.append(token_logprobs)
+            batch_top_logprobs.append(top_logprobs_data.top_logprobs)
+
+            # TODO eventually deprecate
+            # Perplexity = exp(-sum(gold_logprobs)
+            perplexity = np.exp(-1 * np.mean(token_logprobs))
```
bogdan-galileo, line 146: Should this be sum instead of mean ?

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: Okay! Maybe I can even add this as a util function and move out of the seq2seq class?

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: I agree. We also know the length of the list before hand. Can just convert to list.

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: Sure Can do this!

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: I think for perplexity it is mean. Otherwise this would bias against longer sequences.

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -63,103 +66,100 @@ def validate_and_format(self) -> None:
...
-            batch_gold_probs.append(gold_probs)
-        return pa.array(batch_token_deps), batch_deps, pa.array(batch_gold_probs)
+        # Compute the top-k logprob indices across the batch.
+        top_logprob_indices = get_top_logprob_indices(batch_logprobs)
+
+        batch_token_logprobs = []
+        batch_top_logprobs = []
+        batch_perplexities = []
+        # Iterate through the samples in the batch
+        for sample_id, sample_logprobs, sample_top_indices in zip(
```
bogdan-galileo, line 123: Can we vectorize this ? If instead of remove_padding we replace every non-entry with -1000 or a low enough value, can't we vectorize the rest ?

bogdan-galileo: and at the last step we truncate the numpy array of values at each step and convert in pyarrow. Not sure

dataquality/utils/seq2seq.py
jonathangomesselman: This is actually a pretty great call out. I finally was able to get around to doing some timing for this and it seems that this is likely not the way to go about this. 

There are basically 3 ways that I compared. 
1. `tokenizer.decode()`: I originally went for this because it has the benefit of removing strange "tokenizer" specific tokens - e.g. `_` in tokens like `_The`. However, I am actually realizing that it may be nice to have these as this is what the tokenizer is actually doing. 
2. `tokenizer.convert_ids_to_tokens()`: This function converts a token `id` to its "non-formatted" string token - i.e. includes special values like `_`.
3. `tokenizer.vocab`: This method relies on basically flipping the tokenizer vocab from `Map[str: int]` to `Map[int: str]` and then directly accessing this map. The result is the same as `convert_ids_to_tokens`.

Here are the timing results when converting `10_000_000` random token ids (i.e. processing 10 million tokens, which for samples ~100 tokens in length is about 100_000 samples).
image

As we can see there is quite a difference between all three methods, with the current being significantly slower than the other two. For simplicity I might recommend using `convert_ids_to_tokens` since we don't have to store an additional dictionary in our config - though it would be quite simple to invert the map and store it when we store the tokenizer. What do you think @elboy3

bogdan-galileo: Wow nice comparison. How long does `tokernizer.decode` take if you do it by batches of 128 or 512 at a time ?

jonathangomesselman: Doesn't seem to do a ton

image

jonathangomesselman: Move to shortcut ticket for follow up PR

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: Yeah I think this could definitely be a follow up! I know that hf does effectively this with the `attention_mask` they basically do something like:
```
attn_mask = 1 - attn_mask
attn_mask *= -float.inf
logits = logits + attn_mask
```

I'll create a ticket for this as a follow up!

tests/utils/test_seq2seq_utils.py
```diff
@@ -1,8 +1,185 @@
...
+    vocab_size = 100
+    fake_logprobs = np.zeros((seq_len, vocab_size))
+    fake_top_indices = np.zeros((seq_len, 5))
+
+    # We expect labels to have shape (seq_len,) when passing
+    # to process_sample_logprobs
+    incorrect_labels = np.zeros((seq_len, 1))
+
+    with pytest.raises(GalileoException) as excinfo:
+        _, _ = process_sample_logprobs(
```
elboy3, line 135:
```suggestion
        process_sample_logprobs(
```

tests/utils/test_seq2seq_utils.py
```diff
@@ -1,8 +1,185 @@
...
+            fake_logprobs, incorrect_labels, fake_top_indices, mock_tokenizer
+        )
+
+    assert (
+        "Invalid shape (10, 1), process_sample_logprobs"
+        " expects sample_labels to be a 1D array" == str(excinfo.value)
+    )
```
elboy3, line 142: so beast

elboy3 (approved):
🔥 🚀

PR #746, opened 2023-09-20
Feature/generation

jonathangomesselman:

This PR adds generation support for seq2seq models - addressing this shortcut [ticket](https://app.shortcut.com/galileo/story/7661/dq-generation-support-for-encoder-decoder-models).

**Usage**
To support generation in dq, we have started by fully controlling generation internally, guided by the users model and specified generation parameters. We introduce a new `seq2seq.hf` integration, with a watch function to capture:
1. User model
2. GenerationConfig - which is the standard for setting generation parameters with hf
3. Whether the user wants to generate on the training split - Note that this is quite costly.

In code this looks like:
```
model = HF_Model()  // Model with .generate() function
generation_config = GenerationConfig(
    max_new_tokens=15,
    do_sample=True,
    temperature=0.7,
    ....
)

watch(
    model,
    generation_config,
    generate_training_data=False
)
```

When the user calls `.finish()`, behind the scenes we use the user model to generate over the logged data.

**Implementation Details**
Behind the scenes we currently generate over each sample sequentially (i.e. we don't batch process). We have introduced a new key function in `utils/seq2seq.py`: `add_generated_output_to_df`. This is a vaexified function (doesn't materialize the full df into memory) that goes sample by sample to generate the model output.

Here are some notes on the generation process:
1. We leverage the user tokenizer to tokenize the input data and feed this to the model's `.generate()` function. In doing so we keep the model on the GPU
2. From `.generate()` we only capture the generated ids. To get generation log_probs, we run the sample, with generated labels, through the model to get prediction (based on the generation) logits. This gives us a lot more flexibility to handle different generation schemas and frees us from the hidden nature of the `.generate()` function.
3. With the logits and generated ids, we can pretty much treat everything like we are logging data + model outputs as in training!

*Special Note*: To vaexify the generation function - we have to use a pyarrow trick to have a vaex register function return values for multiple columns. Specifically, we take advantage of the `pa.StructArray` and `pa.StructArray.from_arrays`. High level this lets us define basically names 'columns' within a single struct array that we can assign to a single vaex column. Luckily, vaex then supports a `df.struct.flatten` operation to separate the columns in our `StructArray`. **Note** that we take inspiration from a similar implementation in the API [here](https://github.com/rungalileo/api/blob/b69598e1ff1c6bdd8dcdadad9167c4ea198af9ff/api/utils/ner.py#L94).
 
***Open Questions***
@elboy3 One question is whether we want to make a more formal structure for generation rather than having `generation` just as functions in the `utils`. I don't really know the way to think through this, but for example we have `data_loggers` and `model_loggers` - maybe we could have an equivalent `generation_logger`? 

It just feels a bit odd having a bunch of important logic being lumped into the `utils` section.

See test in:
- `tests/loggers/test_seq2seq.py`
- `tests/utils/test_seq2seq_utils.py`


dataquality/schemas/seq2seq.py
```diff
@@ -77,3 +84,12 @@ class LogprobData:
 
     token_logprobs: np.ndarray
     top_logprobs: List[List[Tuple[str, float]]]
+
+
+# TODO Make this work with LogprobData
```
bogdan-galileo, line 89: seems like you made it work ?

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
...
+
+    # Strip the beginning <pad> token and keep as Tensor
+    gen_ids = gen_ids[..., 1:]
+
+    # Pass the generated output through the model to get the logits
+    with torch.no_grad():
+        model_outputs = model(input_ids=input_ids, labels=gen_ids)
+        logits = model_outputs.logits.cpu().numpy()
+
+    # TODO This is much faster on the gpu
```
bogdan-galileo, line 264: so should we just use https://pytorch.org/docs/stable/generated/torch.nn.LogSoftmax.html#torch.nn.LogSoftmax and `.cpu().numpy()` it after ?

tests/utils/test_seq2seq_utils.py
```diff
@@ -1,20 +1,101 @@
...
+    Things to test:
+        - Check that the fake pad token is removed
+        - Check that we have logprobs!
+        - Check that logprobs is correct shape to process_sample_logprobs
+        - Check that gen_ids is correct shape to process_sample_logprobs
+        - Check that we have the correct attributes in ModelGeneration
+    """
+    # Mock the tokenizer
+    mock_tokenizer = mock.MagicMock()
+    mock_tokenizer.return_value = {"input_ids": torch.tensor([[1, 2, 3, 0]])}
```
bogdan-galileo, line 47: One batch of data ? Works with 2 too ?

bogdan-galileo: Oh no, that's just 1 sample ?

tests/utils/test_seq2seq_utils.py
```diff
@@ -1,20 +1,101 @@
...
+
+    # Mock device and generation_config
+    mock_device = torch.device("cpu")
+    mock_generation_config = mock.MagicMock()
+
+    # Mock util helper function. Note we don't mock mock_get_top_logprob_indices
+    # since we only care about the return value
+    fake_token_logprobs = np.array([-0.5, -0.25, -0.11])
+    fake_top_logprob_data = [
+        [("A", -0.1), ("B", -1)],
```
bogdan-galileo, line 70: I dont think it matters, but for fun: the probs of these log probs sum to > 1. Same with the above -0.5, -0.25, -0.11

tests/utils/test_seq2seq_utils.py
```diff
@@ -1,20 +1,101 @@
...
     rollup_offset_mapping,
 )
 
 
+@mock.patch("dataquality.utils.seq2seq.process_sample_logprobs")
+@mock.patch("dataquality.utils.seq2seq.get_top_logprob_indices")
+def test_generate_sample_output(
+    mock_get_top_logprob_indices: mock.Mock, mock_process_sample_logprobs: mock.Mock
+) -> None:
+    """Test the logic for generating over a single sample.
```
bogdan-galileo, line 26: AWesome docstring!! 👍

dataquality/schemas/seq2seq.py
jonathangomesselman: Oops haha this is true!

dataquality/utils/seq2seq.py
jonathangomesselman: Honestly great point! I will do this. I was potentially worried about GPU space? But not sure if this is a real big concern for this operation.

tests/utils/test_seq2seq_utils.py
jonathangomesselman: Yeah for generation right now we just do one sample at a time!

tests/utils/test_seq2seq_utils.py
jonathangomesselman: Hahah good point. But it shouldn't matter since I just check that it gets populated in the return correctly

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -195,6 +197,57 @@ def get_valid_attributes() -> List[str]:
...
+            in_frame, dir_name, prob_only, split, epoch_or_inf
+        )
+
+    @classmethod
+    def add_generated_output_to_df(cls, df: DataFrame, split: str) -> DataFrame:
+        """Adds the generated output to the dataframe
+        Adds the generated output to the dataframe, and also adds the
+        `token_label_positions` column
+        """
+        logger_config = cls.logger_config
```
bogdan-galileo, line 224: I think we should be using `self.logger_config` here otherwise you get the default one, see
Screenshot 2023-09-20 at 3 05 41 PM

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
     return LogprobData(token_logprobs, top_logprobs)
 
 
+def generate_sample_output(
+    input_str: str,
+    model: PreTrainedModel,
+    device: torch.device,
```
bogdan-galileo, line 208: Same about device, maybe we should just infer it from device = model.device() in the code

dataquality/integrations/seq2seq/hf.py
```diff
@@ -0,0 +1,33 @@
...
+def watch(
+    model: PreTrainedModel,
+    generation_config: GenerationConfig,
+    generate_training_data: bool = False,
+) -> None:
+    """Seq2seq only. Log model generations for your run
+
+    Iterates over a given dataset and logs the generations for each sample.
+    `model` must be an instance of transformers PreTrainedModel and have a `generate`
+      method.
```
elboy3, line 18: we should add a small note in the docstring that unlike other `watch` fns we are really just registering the model and generation config, and not actually attaching any hooks or anything to the model. we are just calling it watch for consistency

dataquality/loggers/data_logger/seq2seq.py
elboy3: we should test this. we use `cls.logger_config` all over the place in DQ so i think it still works. 

what does default logger config mean? that we wouldn't have the model. tokenizer that we saved in `watch` ? cause if that were true none of the generation would work, but it's been working

dataquality/schemas/seq2seq.py
```diff
@@ -33,15 +33,15 @@ def set_cols(cls, df: DataFrame) -> DataFrame:
     def set_input(cls, df: DataFrame) -> DataFrame:
         """Sets the input column for the dataframe"""
         if cls.text.value in df.get_column_names():
-            df = df.rename(cls.text.value, cls.input.value)
+            df.rename(cls.text.value, cls.input.value)
```
elboy3, line 36: vaex renames in place, so weird

dataquality/schemas/seq2seq.py
elboy3, line 89: 🧹

dataquality/schemas/seq2seq.py
```diff
@@ -77,3 +84,12 @@ class LogprobData:
...
     top_logprobs: List[List[Tuple[str, float]]]
+
+
+# TODO Make this work with LogprobData
+@dataclass
+class ModelGeneration:
+    generated_ids: np.ndarray
+    generated_logprob_data: LogprobData
+    # generated_token_logprobs: np.ndarray
+    # generated_top_logprobs: List[List[Tuple[str, float]]]
```
elboy3, line 95: 🧹

dataquality/loggers/data_logger/seq2seq.py
bogdan-galileo: Yeah true I see it in the code, we use a mix of `self.logger_config` and `cls.logger_config`.

The default that I was thinking of would be the default value given to it at (class) construction time, here it's `seq2seq_logger_config` that is set to be `seq2seq_logger_config = Seq2SeqLoggerConfig()` if we follow the import.

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -195,6 +197,57 @@ def get_valid_attributes() -> List[str]:
...
+            )
+        if split not in logger_config.generation_splits:
+            print("Skipping generation for split", split)
+            return df
+
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        # Ensure the model is on the correct device and in eval mode
+        model.to(device)
+        model.eval()
+        df = add_generated_output_to_df(df, tokenizer, model, device, generation_config)
```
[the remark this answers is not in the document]
jonathangomesselman: This is a great point. I will remove changing the model device and instead use `device=model.device` to handle the input data for the model during generation!

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
...
+            generation_config=generation_config,
+        )
+
+    # Strip the beginning <pad> token and keep as Tensor
+    gen_ids = gen_ids[..., 1:]
+
+    # Pass the generated output through the model to get the logits
+    with torch.no_grad():
+        model_outputs = model(input_ids=input_ids, labels=gen_ids)
+        logits = model_outputs.logits.cpu().numpy()
```
elboy3, line 262: can this all go in one torch.no_grad ?

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
...
+
+    # Extract out each independent column
+    df = df.struct.flatten(C.generation_data.value)
+    # struct.flatten creates a column per key (created above),
+    # where the column name will be `Combined_Output_<key>` so we rename them
+    for col in generated_columns:
+        df.rename(f"{C.generation_data.value}_{col}", col)
+
+    # After flattening vaex pre-pends 4 `_`s
+    df = df.drop(f"____{C.generation_data.value}")
```
elboy3, line 398: so weird. does it keep `generation_data` at all?

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
...
+    columns.
+    """
+
+    generated_columns = [
+        C.generated_output.value,
+        C.generated_token_label_positions.value,
+        C.generated_token_label_offsets.value,
+        C.generated_token_logprobs.value,
+        C.generated_top_logprobs.value,
+    ]
```
elboy3, line 321: this could be a list as a property returned by the class `Seq2SeqOutputCols`

```
@property
def generation_cols() -> List[str]:
...
```

dataquality/utils/seq2seq.py
```diff
@@ -191,6 +202,203 @@ def process_sample_logprobs(
...
+        generated_token_label_offsets = []
+        generated_token_logprobs = []
+        generated_top_logprobs = []
+
+        for sample in texts:
+            # Generate and extract model outputs
+            sample_generation = generate_sample_output(
+                input_str=str(sample),
+                model=model,
+                device=device,
```
elboy3, line 345: i kinda agree not having to pass device through each fn is nice and inferring it from model would be clean, but im less familiar with torch patterns so if that's normal we can keep it

dataquality/loggers/data_logger/seq2seq.py
jonathangomesselman: Hmmm I am not sure I fully follow this. Haha I am with @elboy3 that it seems to be working?

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: Okay! Will add

dataquality/schemas/seq2seq.py
jonathangomesselman: Yeah I noticed this. I got an error that was like we're trying to index a 'str' named 'text'.

dataquality/utils/seq2seq.py
jonathangomesselman: Good point haha!

dataquality/utils/seq2seq.py
jonathangomesselman: Okay so I thought I had this figured out, but the layers keep getting deeper. After `df = df.struct.flatten(C.generation_data.value)` the df columns look like this:
```
['text',
 'generation_data_generated_output',
 'generation_data_generated_token_label_positions',
 'generation_data_generated_token_label_offsets',
 'generation_data_generated_token_logprobs',
 'generation_data_generated_top_logprobs',
 '____generation_data']
```
As you can see it weirdly keep `generation_data` but adds 4 "_"s. 

What is then mega bizarre is after renaming the other columns and then trying to trop ` ____generation_data` look what happens:
```
df.column_names -> ['text',
 'generated_output',
 'generated_token_label_positions',
 'generated_token_label_offsets',
 'generated_token_logprobs',
 'generated_top_logprobs',
 '______generation_data']
```
It creates a new column now with 6 "_"s???!! Wtf vaex I am so confused. Any ideas here. I guess we could always slice on just the columns we want, but drop feels cleaner.

dataquality/utils/seq2seq.py
jonathangomesselman: Oh fancy!

dataquality/utils/seq2seq.py
jonathangomesselman: I like @bogdan-galileo 's suggestion. I changed to that!

dataquality/utils/seq2seq.py
jonathangomesselman: Maybe you can show me how to do this, because I am not sure exactly how to do it without instantiating the class as an object?

dataquality/loggers/data_logger/seq2seq.py
setu4993: +1 to what Bogdan is saying here. This is a case of class defaults (`cls.attribute`) and instantiated values (`self.attribute`).

I think what Bogdan is saying that we want the instantiated value but using cls gives us the default value always, which will be incorrect.

dataquality/schemas/seq2seq.py
```diff
@@ -51,6 +51,13 @@ class Seq2SeqOutputCols(str, Enum):
     perplexity = "perplexity"
     token_logprobs = "token_logprobs"
     top_logprobs = "top_logprobs"
+    # Columns associated with generated output
+    generated_output = "generated_output"
+    generated_token_label_positions = "generated_token_label_positions"
+    generated_token_label_offsets = "generated_token_label_offsets"
+    generated_token_logprobs = "generated_token_logprobs"
+    generated_top_logprobs = "generated_top_logprobs"
+    generation_data = "generation_data"
```
elboy3, line 60: we aren't saving this to the df so im not sure it's needed in outputcols

dataquality/integrations/seq2seq/hf.py
```diff
@@ -0,0 +1,62 @@
...
+    set_tokenizer(tokenizer)
+
+    seq2seq_logger_config.model = model
+    seq2seq_logger_config.generation_config = generation_config
+
+    generation_splits = {Split.validation, Split.test}
+    if generate_training_data:
+        generation_splits.add(Split.training)
+
+    seq2seq_logger_config.generation_splits = generation_splits
```
bogdan-galileo, line 62: I don't think that's gonna work ? In here we change the object `seq2seq_logger_config` which is a specific instance that is created in `dataquality.loggers.logger_config.seq2seq`. Which is not the same instance as the one that is the attribute of the data logger.

bogdan-galileo: I'm testing with 
```
watch(
    model,
    tokenizer,
    generation_config,
    generate_training_data=GENERATE_ON_TRAIN,
)
```
and still getting the error 
`AssertionError: You must set your tokenizer before logging. Use dq.set_tokenizer`
like we never set the tokenizer

jonathangomesselman: @elboy3 ?

dataquality/utils/seq2seq.py
jonathangomesselman: `_` Columns seem to be hidden columns in vaex. For now let's ignore this until we see some error come up related to it.

elboy3 (approved):
🚀 🚀

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: @bogdan-galileo I'm not seeing this issue?

PR #747, opened 2023-09-20
Allow changing max num tokens (input and output)

bogdan-galileo:
Adding support for max_input_tokens and max_output_tokens. Knowing these values is necessary for a few reasons:
1. for both of them: we want to indicate in the UI which part of the input/output was not used
2. for output: we currently tokenize the entire output, knowing where teaching forcing stops would allow us to tokenize less (speed improvement)

So in this PR we
- fix DQ when the tokenizer truncates the target (currently indexing error as it expects all the tokens to be there)
- add a column in the df indicating how much of the input went through the model --> for UI
- add a column in the df indicating how much of the target was used by the model (teacher forcing) --> for UI


bogdan-galileo:
Split into smaller PRs https://github.com/rungalileo/dataquality/pull/750 and https://github.com/rungalileo/dataquality/pull/751

PR #750, opened 2023-09-22
Save max input token lengths

bogdan-galileo:
>
>

***Demonstration***

### DIFF
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -1,3 +1,6 @@
+from typing import Optional
+from warnings import warn
+
 from transformers import GenerationConfig, PreTrainedModel, PreTrainedTokenizerFast
 
 from dataquality.loggers.logger_config.seq2seq import seq2seq_logger_config
@@ -8,7 +11,11 @@ from dataquality.utils.task_helpers import get_task_type
 
 
 @check_noop
-def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
+def set_tokenizer(
+    tokenizer: PreTrainedTokenizerFast,
+    max_input_tokens: Optional[int] = None,
+    max_target_tokens: Optional[int] = None,
+) -> None:
     """Seq2seq only. Set the tokenizer for your run
 
     Must be a fast tokenizer, and must support `decode`, `encode`, `encode_plus`
@@ -22,6 +29,28 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
     seq2seq_logger_config.tokenizer = tokenizer
+
+    if max_input_tokens is not None:
+        seq2seq_logger_config.max_input_tokens = max_input_tokens
+    else:
+        seq2seq_logger_config.max_input_tokens = tokenizer.model_max_length
+        warn(
+            (
+                "The argument max_input_tokens is not set, we will use the value "
+                "found in tokenizer.model_max_length. If you tokenized the input with "
+                "another value, this can lead to confusing insights about this "
+                "training run."
+            )
+        )
+
+    seq2seq_logger_config.max_target_tokens = max_target_tokens
+    if seq2seq_logger_config.max_target_tokens is None:
+        warn(
+            (
+                "The argument max_target_tokens is not set. Passing it as an argument "
+                "will speed up logging of the dataset."
+            )
+        )
     # Seq2Seq doesn't have labels but we need to set this to avoid validation errors
     seq2seq_logger_config.labels = []
 
@@ -32,6 +61,8 @@ def watch(
     tokenizer: PreTrainedTokenizerFast,
     generation_config: GenerationConfig,
     generate_training_data: bool = False,
+    max_input_tokens: Optional[int] = None,
+    max_target_tokens: Optional[int] = None,
 ) -> None:
     """Seq2seq only. Log model generations for your run
 
@@ -50,7 +81,7 @@ def watch(
     ), "model must be an instance of transformers PreTrainedModel"
     assert model.can_generate(), "model must contain a `generate` method for seq2seq"
 
-    set_tokenizer(tokenizer)
+    set_tokenizer(tokenizer, max_input_tokens, max_target_tokens)
 
     seq2seq_logger_config.model = model
     seq2seq_logger_config.generation_config = generation_config
--- a/dataquality/loggers/data_logger/seq2seq.py
+++ b/dataquality/loggers/data_logger/seq2seq.py
@@ -109,8 +109,15 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
             "You must set your tokenizer before logging. "
             "Use `dq.integrations.seq2seq.hf.set_tokenizer`"
         )
+        # We set verbose=False: if max_length was not specified by the user and is None,
+        # the tokenizer will throw a warning that is irrelevant in this case (as it uses
+        # max_length=max_model_length, which is the value for the input, not the output)
         encoded_data = self.logger_config.tokenizer(
-            self.labels, return_offsets_mapping=True
+            self.labels,
+            return_offsets_mapping=True,
+            max_length=self.logger_config.max_target_tokens,
+            truncation=False if self.logger_config.max_target_tokens is None else True,
+            verbose=False,
         )
         self.tokenized_labels = encoded_data["input_ids"]
         aligned_data = align_tokens_to_character_spans(encoded_data["offset_mapping"])
@@ -228,6 +235,7 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
         logger_config = cls.logger_config
         model = logger_config.model
         tokenizer = logger_config.tokenizer
+        max_input_tokens = logger_config.max_input_tokens
         generation_config = logger_config.generation_config
         if model is None:
             raise GalileoException(
@@ -239,6 +247,7 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
                 "You must set your tokenizer before logging. Use "
                 "`dataquality.integrations.seq2seq.hf.watch`"
             )
+        assert isinstance(max_input_tokens, int)
         if generation_config is None:
             raise GalileoException(
                 "You must set your generation config before logging. Use "
@@ -248,7 +257,9 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
             print("Skipping generation for split", split)
             return df
 
-        df = add_generated_output_to_df(df, model, tokenizer, generation_config)
+        df = add_generated_output_to_df(
+            df, model, tokenizer, max_input_tokens, generation_config
+        )
         return df
 
     @classmethod
--- a/dataquality/loggers/model_logger/seq2seq.py
+++ b/dataquality/loggers/model_logger/seq2seq.py
@@ -130,7 +130,10 @@ class Seq2SeqModelLogger(BaseGalileoModelLogger):
         for sample_id, sample_logprobs, sample_top_indices in zip(
             batch_ids, batch_logprobs, top_logprob_indices
         ):
-            sample_labels = self._retrieve_sample_labels(sample_id)
+            sample_n_tokens = sample_logprobs.shape[0]
+            sample_labels = self._retrieve_sample_labels(
+                sample_id, max_tokens=sample_n_tokens
+            )
             padding_side = self.logger_config.tokenizer.padding_side
             sample_logprobs = remove_padding(
                 sample_labels,
@@ -180,18 +183,20 @@ class Seq2SeqModelLogger(BaseGalileoModelLogger):
     def _write_dict_to_disk(self, path: str, object_name: str, data: Dict) -> None:
         save_arrow_file(path, object_name, data)
 
-    def _retrieve_sample_labels(self, sample_id: int) -> np.ndarray:
-        """Retrieve the labels array based on the sample id
+    def _retrieve_sample_labels(
+        self, sample_id: int, max_tokens: Optional[int] = None
+    ) -> np.ndarray:
+        """Retrieve the labels array based on the sample id and truncate at max_tokens
 
         Labels gives the ground truth / target sample ids for
         each token in the sequence:
 
         e.g. for sample_id = 8 --> labels = [0, 10, 16, ...]
         """
-        labels = np.array(
-            self.logger_config.id_to_tokens[self.token_map_key][sample_id]
-        )
-        return labels
+        labels = self.logger_config.id_to_tokens[self.token_map_key][sample_id]
+        if max_tokens is not None:
+            labels = labels[:max_tokens]
+        return np.array(labels)
 
     def convert_logits_to_logprobs(
         self, sample_logits: Union[List[np.ndarray], np.ndarray]
--- a/tests/loggers/test_seq2seq.py
+++ b/tests/loggers/test_seq2seq.py
@@ -198,7 +198,7 @@ def test_add_generated_output_to_df(
 ) -> None:
     """Test the complex vaex batched processing function for generation
 
-    Tings to Mock
+    Things to Mock
         - generate_sample_output: This can be fairly simple. The
         one thing that we want to vary would be the length of things returned
         - tokenizer: decode can be quite simple + encode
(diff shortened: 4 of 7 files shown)


dataquality/integrations/seq2seq/hf.py
```diff
@@ -22,6 +29,28 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
     seq2seq_logger_config.tokenizer = tokenizer
+
+    if max_input_tokens is not None:
+        seq2seq_logger_config.max_input_tokens = max_input_tokens
+    else:
+        seq2seq_logger_config.max_input_tokens = tokenizer.model_max_length
```
elboy3, line 36: is `tokenizer.model_max_length` guaranteed to be set?

dataquality/integrations/seq2seq/hf.py
```diff
@@ -22,6 +29,28 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
...
+        warn(
+            (
+                "The argument max_input_tokens is not set, we will use the value "
+                "found in tokenizer.model_max_length. If you tokenized the input with "
```
elboy3, line 40: maybe f-print the value used as well?

dataquality/integrations/seq2seq/hf.py
```diff
@@ -8,7 +11,11 @@
 
 
 @check_noop
-def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
+def set_tokenizer(
+    tokenizer: PreTrainedTokenizerFast,
+    max_input_tokens: Optional[int] = None,
+    max_target_tokens: Optional[int] = None,
+) -> None:
```
elboy3, line 18: update docstring with info on max_input_tokens and max_target_tokens

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: For T5 at least yeah, it has a default value of 512

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -109,8 +109,15 @@ def validate_and_format(self) -> None:
             "You must set your tokenizer before logging. "
             "Use `dq.integrations.seq2seq.hf.set_tokenizer`"
         )
+        # We set verbose=False: if max_length was not specified by the user and is None,
```
elboy3, line 112: it looks like verbose is always False though?

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -109,8 +109,15 @@ def validate_and_format(self) -> None:
...
+        # the tokenizer will throw a warning that is irrelevant in this case (as it uses
+        # max_length=max_model_length, which is the value for the input, not the output)
         encoded_data = self.logger_config.tokenizer(
-            self.labels, return_offsets_mapping=True
+            self.labels,
+            return_offsets_mapping=True,
+            max_length=self.logger_config.max_target_tokens,
+            truncation=False if self.logger_config.max_target_tokens is None else True,
```
elboy3, line 119:
```suggestion
            truncation=(self.logger_config.max_target_tokens is not None),
```

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: Actually it is set for all tokenizer.

model_max_length (int, optional) — The maximum length (in number of tokens) for the inputs to the transformer model. When the tokenizer is loaded with [from_pretrained()](https://huggingface.co/docs/transformers/v4.33.2/en/internal/tokenization_utils#transformers.PreTrainedTokenizerBase.from_pretrained), this will be set to the value stored for the associated model in max_model_input_sizes (see above). If no value is provided, will default to VERY_LARGE_INTEGER (int(1e30)).

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -130,7 +130,10 @@ def process_logprobs(
         for sample_id, sample_logprobs, sample_top_indices in zip(
             batch_ids, batch_logprobs, top_logprob_indices
         ):
-            sample_labels = self._retrieve_sample_labels(sample_id)
+            sample_n_tokens = sample_logprobs.shape[0]
+            sample_labels = self._retrieve_sample_labels(
+                sample_id, max_tokens=sample_n_tokens
+            )
```
elboy3, line 136: why is this max_tokens instead of the val set in config?

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: done

elboy3 (commented):
nice work! left some comments and questions. can we also add a test where max length is less than the max length of tokens in a sample and assert that it was truncated ? 

also aren't we going to add length to the df? i don't think i see that done here

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: Haha if no value is set that will be an awkward situation if it is 1e30.

dataquality/loggers/data_logger/seq2seq.py
bogdan-galileo: Yeah. The only comment that I ever saw was that one

`Token indices sequence length is longer than the specified maximum sequence length for this model (877 > 512). Running this sequence through the model will result in indexing errors`

We could check and set it True if we're not in that scenario, but I never saw another message and I'm not sure it's of much value

dataquality/integrations/seq2seq/hf.py
```diff
@@ -22,6 +29,28 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
...
+        warn(
+            (
+                "The argument max_input_tokens is not set, we will use the value "
+                "found in tokenizer.model_max_length. If you tokenized the input with "
+                "another value, this can lead to confusing insights about this "
+                "training run."
+            )
+        )
+
+    seq2seq_logger_config.max_target_tokens = max_target_tokens
```
jonathangomesselman, line 46: Is there a reason we use `model.max_model_length` for just the inputs? Is this commonly how it is done. I haven't looked for example at Encoder-Decoder models, if they will have different pre-set values for the Encoder and Decoder block.

dataquality/loggers/data_logger/seq2seq.py
bogdan-galileo: I did that at first but then you run into 
Screenshot 2023-09-22 at 2 26 24 PM

I think we can make it work with != None, but to be safe I just wrote it all

jonathangomesselman: Do we ever want `truncation` to be `False`? Wouldn't we want to at least always truncate to the max token length of the model set in the `tokenizer` by default? I guess I don't see why we would ever want to tokenizer say a full text with 1500 tokens if the model will reasonably never output this.

dataquality/loggers/model_logger/seq2seq.py
jonathangomesselman: @elboy3 This is specifically to catch the case where the choose or forget to set a `max_target_length` and then use a smaller value than the default. Previously this would cause an error (because we later try to index for more labels than exist). But now we just truncate the labels to whatever we see the user has truncated to.

dataquality/loggers/model_logger/seq2seq.py
```diff
@@ -180,18 +183,20 @@ def _get_data_dict(self) -> Dict:
     def _write_dict_to_disk(self, path: str, object_name: str, data: Dict) -> None:
         save_arrow_file(path, object_name, data)
 
-    def _retrieve_sample_labels(self, sample_id: int) -> np.ndarray:
-        """Retrieve the labels array based on the sample id
+    def _retrieve_sample_labels(
+        self, sample_id: int, max_tokens: Optional[int] = None
```
jonathangomesselman, line 187: Is there a reason this is optional? 

Maybe we can also update the doc string to mention why we have this variable!

tests/loggers/test_seq2seq.py
```diff
@@ -198,7 +198,7 @@ def test_add_generated_output_to_df(
 ) -> None:
```
jonathangomesselman, line 198: This may be un-necessary, but it seems nice to have a test to validate that we can in fact change the max_token length. I see a couple of cases to check :
1. When user provides max length for target make sure the labels are properly tokenized on our end.
2. The user only provides max_input - check this is properly used in generation.
3. The user does not log target_max_length - make sure that the labels are properly truncated when getting logprobs.

jonathangomesselman (commented):
Great stuff. Left also a couple of comments. 

My main question is around why we use different logic `max_input_length` and `max_target_length`. For `max_input_length` we default to the model max_length, whereas for `max_target_length` we default to `None`. Moreover, when then tokenizing we don't truncate if they don't provide `max_target_length`. This feels odd since in general the tokenizer defaults to having the model_max_length and we would never need more tokens then that.

I guess as I write this I am curious if you did this to allow for the most "flexability" / "safety." Essentially, we don't make any assumptions about the Targets until we actually see the logits come in? My only question is whether this can actually lead to issues?

Second thing is whether we want to have 1 or 2 tests to just make sure the logic works for these. I wrote I think 3 cases that we could watch for / are affected. I think in general these could be simple, but maybe worth having quick tests for.

dataquality/loggers/model_logger/seq2seq.py
bogdan-galileo: Before this we already tokenized the output truncating at `max_target_length` (if it's None, we don't truncate). Right here we load these tokens and compare them with the model output.

The extra truncation that I do here is `our_tokens = our_tokens[:size_of_their_tokens]`. If `our_tokens > size_of_their_tokens` we get an index error, so it's helpful to have it

I think I would keep it, here is a scenario where we get that error
- they load a vanilla tokenizer
- they tokenize their output with truncate=True, it uses the default 512 value
- they log the tokenizer but no max_target_length
- they didn't log so we assume they don't tokenize. But they did and it breaks.

Thoughts?

bogdan-galileo: Or Jon's case. Thanks, easier... lol

dataquality/loggers/model_logger/seq2seq.py
bogdan-galileo: Good point, I thought if we re-use it, but we defintely can't reuse that method for anything else. Will update the doc!

dataquality/loggers/data_logger/seq2seq.py
bogdan-galileo: They do get a warning that says that logging would be sped up if they provide that value.

But mostly it's that I haven't seen a max_token_length in the tokenizer for the OUTPUT. All I saw was

`model_max_length` (int, optional) — The maximum length (in number of tokens) for the inputs to the transformer model...

`max_model_input_sizes` (Dict[str, Optional[int]]) — A dictionary with, as keys, the short-cut-names of the pretrained models, and as associated values, the maximum length of the sequence inputs of this model, or None if the model has no maximum input size.

There is also `max_len_sentences_pair`, `max_len_single_sentence`

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: Yeah from the doc seems like it's used for the input.

model_max_length (int, optional) — The maximum length (in number of tokens) for the inputs to the transformer model. When the tokenizer is loaded with from_pretrained(), this will be set to the value stored for the associated model in max_model_input_sizes (see above). If no value is provided, will default to VERY_LARGE_INTEGER (int(1e30)).

I guess for the output it could be unbounded ? I m not sure

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: Added

dataquality/loggers/data_logger/seq2seq.py
jonathangomesselman: I think that in general `model_max_length` is expected to be the same for both the encoder and decoder by default? I am not 100% sure about this, but I think that is the general approach? 

Take for example machine translation. The input and outputs generally have a similar number of tokens. So I think that this assumption / design makes pretty reasonable sense.

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: I think that "input" is unfortunately an overloaded / poorly defined term. We use Input to specifically refer to the inputs to the Decoder. However, I think they more generally refer to input to either the Encoder or Decoder. My intuition for this is because the variable names to each are:
- input_ids
- decoder_input_ids
So even for the decoder they are reffered to as `input_ids`.

dataquality/loggers/data_logger/seq2seq.py
bogdan-galileo: removing this since we will always truncate now

tests/loggers/test_seq2seq.py
bogdan-galileo: Just added!

dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -80,8 +80,6 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
 
     def __init__(self, meta: Optional[MetasType] = None) -> None:
         super().__init__(meta)
-        # Tokens IDs in a given input string
-        self.tokenized_labels: List[List[int]] = []
```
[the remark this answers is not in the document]
jonathangomesselman: We just need to save it right?

dataquality/integrations/seq2seq/hf.py
```diff
@@ -22,6 +33,31 @@ def set_tokenizer(tokenizer: PreTrainedTokenizerFast) -> None:
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
     seq2seq_logger_config.tokenizer = tokenizer
+
+    if max_input_tokens is not None:
```
jonathangomesselman, line 37: Super tiny nit, but for consistency do we want to have the check for `max_input_tokens` and for `max_target_tokens` be the same? And for this I literally just mean the:
```
seq2seq_logger_config.max... = max...
if seq2seq_logger_config.max... is None:
  # Do stuff
```

jonathangomesselman (approved):
Looks good to me!! 🚢 will defer to @elboy3 the pytest master :)

dataquality/loggers/data_logger/seq2seq.py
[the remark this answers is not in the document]
bogdan-galileo: Yeah as a local variable in that method. I didn't see it used anywhere else!

dataquality/integrations/seq2seq/hf.py
bogdan-galileo: whoops, copy pasted wrong. Thanks for catching that!

dataquality/loggers/data_logger/seq2seq.py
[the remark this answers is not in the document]
elboy3: classic ben

tests/loggers/test_seq2seq.py
```diff
@@ -274,3 +279,123 @@ def mock_generate_output() -> ModelGeneration:
...
+        generate_sample_output(
+            input_text,
+            mock_model,
+            tokenizer_T5,
+            seq2seq_logger_config.max_input_tokens,
+            mock_generation_config,
+        )
+    except TypeError as e:
+        # Assert that we exit because model was mocked
+        assert str(e) == "'Mock' object is not subscriptable"
```
elboy3, line 305: the pytest way to do this is 

```python
    with pytest.raises(TypeError) as e:
        generate_sample_output(
            input_text,
            mock_model,
            tokenizer_T5,
            seq2seq_logger_config.max_input_tokens,
            mock_generation_config,
        )
```

elboy3: also it seems weird that we are asserting a mock in a test, maybe we sync on this briefly?

tests/loggers/conftest.py
```diff
@@ -62,3 +63,17 @@ def curry(
...
+    return torch.tensor([[1, 2, 3], [4, 5, 6]])
+
+
+class ModelOutput:
+    logits = torch.tensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
+
+
+@pytest.fixture
+def seq2seq_model_outputs() -> ModelOutput:
+    return ModelOutput()
```
elboy3, line 79:
```suggestion
@pytest.fixture
def seq2seq_model_outputs() -> MagicMock:
    return MagicMock(
        logits = torch.tensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
    )
```

elboy3: i think something like this would work too. test it before merging! it'd be more pytest-y though :)

elboy3 (approved):

PR #753, opened 2023-09-27
Temp/fix generation

jonathangomesselman:
### Description
We are using a vaex register function to do generation. This is nice because we can take advantage of vaex batched processing to avoid bringing everything into memory - here the "input" column. Moreover, a virtual column is created to "lazily" compute the columns as needed.

While virtual columns are great in certain scenarios, they are not so great in this one. Virtual columns seem to be especially useful when we are working with HUGE dfs and want to avoid at all cost bringing tons of data into memory. The trade-off is that virtual columns are not "materialized" in memory, so anytime we need to access them we have to re-compute. This is BAD for us since generation is a very slow process (that honestly does not take up all that much memory for the small datasets that people generate on). 

**Bonus issue**: it seems that the `flatten` operation in vaex leads to some unstable batching behavior and accounts for quite a bit of duplicate compute.

### Solution
Use `df = df.materialize('col')`. Basically, we want to just force the computation of the generation columns and bring them into memory (while still batching over the `Input` column). This way we avoid ever having to recompute them. BONUS: we can also avoid the strange behavior of `flatten` by materializing the Struct column before flattening. 

See as reference where we do this similar tradeoff when adding `pca_emb` to the df. A similar idea is used there - we sacrifice the extra storage of brining in the full `pca_emb` (while still avoiding brining in the `full_embs`) in favor of avoiding the costly process of recomputing. See [github](https://github.com/rungalileo/dataquality/blob/fb76b5aa629687d2acfde599bf12137fb230ff7c/dataquality/utils/vaex.py#L288)


jonathangomesselman:
@elboy3 Ugh whoops. Fixed that!

PR #754, opened 2023-09-28
fix: generation without vaex

jonathangomesselman:

This PR address this [ticket](https://app.shortcut.com/galileo/story/7910)

***Bug Description***
Currently generation works with small dataset sizes, but fails when we try over > 500 samples. We get some very strange error related to the `StructArray` that we have to use. When I talked to Ben about this, he mentioned that while possible this vaex `StructArray` design was fairly temperamental and error prone - which it seems a might be. 

***Proposed Solution***
The primary reason to use a vaex register function is to avoid bringing large amounts of data into memory through lazy evaluation. However, as addressed by a previous issue, we end up effectively forcing the computation of the `generated_data` to avoid lazy recomputation. Therefore, we are not really using the vaex register function correctly.

With this in mind, we propose switching to a more manual batched processing of the generation job. This will give us more control over generation, especially around potential errors. We basically move to a manual batched approach using vaex's `.evaluate_iterator()`.

We introduce a new DataClass type called `BatchGenerationData`. This data class stores all the information needed for generation. Moreover, for batched processing we allow for initializing an empty central `generation_data` object, followed by easy *in-place* extending with a new batch of generation data. See this example:
```
generation_data = BatchGenerationData.empty_init()
for _ in range(num_batches):
  batch_generated_data = ...
  # Add new data to central data store
  generation_data.extend_from(batch_generated_data)
...
```

**Note**: We favor in-place adding of data for memory and performance concerns.

### DIFF
--- a/dataquality/schemas/seq2seq.py
+++ b/dataquality/schemas/seq2seq.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from enum import Enum
 from typing import List, Set, Tuple
 
@@ -125,6 +125,7 @@ class BatchGenerationData:
     top_logprobs: List[List[List[Tuple[str, float]]]]
         top_logprobs for each sample
     """
+
     generated_outputs: List[str]
     generated_token_label_positions: List[List[Set[int]]]
     generated_token_label_offsets: List[List[Tuple[int, int]]]
@@ -132,14 +133,14 @@ class BatchGenerationData:
     generated_top_logprobs: List[List[List[Tuple[str, float]]]]
 
     @classmethod
-    def empty_init(cls) -> 'BatchGenerationData':
+    def empty_init(cls) -> "BatchGenerationData":
         """Initialize a BatchGenerationData object with no sample data"""
         return cls(
             generated_outputs=[],
             generated_token_label_positions=[],
             generated_token_label_offsets=[],
             generated_token_logprobs=[],
-            generated_top_logprobs=[]
+            generated_top_logprobs=[],
         )
 
     def extend_from(self, batch_data: "BatchGenerationData") -> None:
@@ -149,7 +150,11 @@ class BatchGenerationData:
         memory and performance.
         """
         self.generated_outputs.extend(batch_data.generated_outputs)
-        self.generated_token_label_positions.extend(batch_data.generated_token_label_positions)
-        self.generated_token_label_offsets.extend(batch_data.generated_token_label_offsets)
+        self.generated_token_label_positions.extend(
+            batch_data.generated_token_label_positions
+        )
+        self.generated_token_label_offsets.extend(
+            batch_data.generated_token_label_offsets
+        )
         self.generated_token_logprobs.extend(batch_data.generated_token_logprobs)
         self.generated_top_logprobs.extend(batch_data.generated_top_logprobs)
--- a/dataquality/schemas/seq2seq.py
+++ b/dataquality/schemas/seq2seq.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import List, Set, Tuple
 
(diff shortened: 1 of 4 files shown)


jonathangomesselman:
@bogdan-galileo @elboy3 - Still need to add some updated tests. 

Additionally, @bogdan-galileo maybe you or I could test that this actually solves our issues haha! 

Overall putting this here for us to get eyes on it and test it out 🚀

jonathangomesselman:
I have added tests that should cover everything added!

jonathangomesselman:
@elboy3 Raised also a good point. Right now we still use the batched processing of vaex to batch generation. The only real gain that we get for this is avoiding bringing in the full INPUT text column - then at the same time we generate and store all of the generation data in memory. 

So in reality we are really not saving all that much ...

Elliott's idea was to maybe just ditch the batched processing and process the full INPUT column as a single batch. I am not super opinionated about either. The batching seems to work right now, so maybe we keep it, but also it would be very easy to remove it! Curious people's thoughts.

dataquality/schemas/seq2seq.py
```diff
@@ -103,3 +105,60 @@ class LogprobData:
...
+        Token label logprobs for each sample
+    top_logprobs: List[List[List[Tuple[str, float]]]]
+        top_logprobs for each sample
+    """
+
+    generated_outputs: List[str]
+    generated_token_label_positions: List[List[Set[int]]]
+    generated_token_label_offsets: List[List[Tuple[int, int]]]
+    generated_token_logprobs: List[np.ndarray]
+    generated_top_logprobs: List[List[List[Tuple[str, float]]]]
```
elboy3, line 137:
```suggestion
    generated_outputs: List[str] = field(default_factory=list)
    generated_token_label_positions: List[List[Set[int]]] = field(default_factory=list)
    generated_token_label_offsets: List[List[Tuple[int, int]]] = field(default_factory=list)
    generated_token_logprobs: List[np.ndarray] = field(default_factory=list)
    generated_top_logprobs: List[List[List[Tuple[str, float]]]] = field(default_factory=list)
```

elboy3: and then you'll need to import `field` from dataclasses above

jonathangomesselman: Intersting! I was going to do this but decided against it. Do you prefer this over the empty init?

jonathangomesselman: GPT-4 wouldn't tell me which was nicer lol

elboy3:
https://app.shortcut.com/galileo/story/7910/dq-remove-vaex-register-function-in-generation

dataquality/utils/seq2seq/generation.py
```diff
@@ -137,81 +208,30 @@ def add_generated_output_to_df(
...
+    )
+    df[S2SOC.generated_token_label_offsets.value] = pa.array(
+        generated_data.generated_token_label_offsets
+    )
+    df[S2SOC.generated_token_logprobs.value] = pa.array(
+        generated_data.generated_token_logprobs
+    )
+    df[S2SOC.generated_top_logprobs.value] = pa.array(
+        generated_data.generated_top_logprobs, type=TOP_LOGPROBS_SCHEMA
+    )
```
elboy3, line 235: this is so much cleaner / readable than the vaex struct thing we were doing before 🙌

jonathangomesselman: Yeah it really is!

dataquality/utils/seq2seq/generation.py
```diff
@@ -117,9 +187,10 @@ def add_generated_output_to_df(
         - generated_token_logprobs: pa.array
         - generated_top_logprobs: pa.array
 
-    Note: We return a pa.StructArray to extract multiple columns of info
-    at once through vaex. We then have to seperate the Struct into individual
-    columns.
+    NOTE: Although we bring into memory quite a bit of information about the generated
+    outputs, in general users won't be generated over very many samples (on the order of
+    100-1000s because it simply takes too much time to do much more). Nevertheless, we
```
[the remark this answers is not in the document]
jonathangomesselman: That makes sense! I figured if we really were pressed to avoid memory we could maybe do something like this - but was not exactly sure how to do it.

elboy3 (approved):
🚀  awesome work!

jonathangomesselman:
@elboy3 Final step is to do a test run with a big generation dataset. Will do that today!

jonathangomesselman:
@elboy3 - Results of small speed test between `main` and this branch:

For this comparison I use the following model hyper-parameters:
- Model: Flan-T5-Large
- Max_input_tokens: 128
- Max_gen_tokens: 16
- GPU: V100

I then test generation on a dataset with **500** Samples

***Baseline main***
```
time-finish() = 368 sec = ~6 minutes (500 samples)

***This branch***
```
time-finish() = 366 sec =  ~6 minutes (500 samples)
```

Overall we get shockingly similar performances - which I think is somewhat as expected. The majority of processing time is split between model generation and output processing. So overall it is a win that thing didn't slow down for some reason!

jonathangomesselman:
https://app.shortcut.com/galileo/story/7943

PR #755, opened 2023-09-29
Bug/gen empty output

jonathangomesselman:

Shortcut ticket: [link](https://app.shortcut.com/galileo/epic/5806?cf_workflow=500000005&ct_workflow=all&group_by=none&vc_group_by=day)

***Bug Description***
DQ logging fails when handling sequences of length one. This can happen for example in the case of an empty string with T5 where T5 represents the empty string as `input_ids = [1]` (i.e. just the `[EOS]` token). So e.g. during generation the model may immediately generate the `[EOS]` token leading, leading to a single token sequence.

***Root cause analysis***
This bug is a result of using the `.squeeze()` function without properly specifying dimensions. We use `.squeeze()` to remove singleton e.g. batch dimensions; however, when we have `seq_len=1`, `.squeeze()` will also remove this sequence dimension. See this example:
```
gen_ids = gen_ids[..., 1:] # Shape - (1, seq_len)
....
gen_ids = gen_ids.squeeze().cpu().numpy() # [seq_len]
```
If `seq_len = 1` then `gen_ids` will have  `shape = ()` rather than `(seq_len)`, since `squeeze()` removes all singleton dimensions. 

The fix that we add is that whenever we use squeeze we specify which dimension we specifically are squeezing! E.g.
```
gen_ids = gen_ids[..., 1:] # Shape - (1, seq_len)
....
# Squeeze the batch dimension
gen_ids = gen_ids.squeeze(0).cpu().numpy() # [seq_len]
```

### DIFF
--- a/tests/utils/test_seq2seq_utils.py
+++ b/tests/utils/test_seq2seq_utils.py
@@ -104,9 +104,7 @@ def test_generate_sample_output(
 
 
 @mock.patch("dataquality.utils.seq2seq.generation.get_top_logprob_indices")
-def test_generate_sample_output_empty_sample(
-    mock_get_top_logprob_indices: mock.Mock
-):
+def test_generate_sample_output_empty_sample(mock_get_top_logprob_indices: mock.Mock):
     # Mock the tokenizer
     mock_tokenizer = mock.MagicMock()
     mock_tokenizer.return_value = {"input_ids": torch.tensor([[1, 2, 3, 0]])}
--- a/tests/utils/test_seq2seq_utils.py
+++ b/tests/utils/test_seq2seq_utils.py
@@ -103,6 +103,45 @@ def test_generate_sample_output(
     assert model_generation.generated_logprob_data.top_logprobs == fake_top_logprob_data
 
 
+@mock.patch("dataquality.utils.seq2seq.generation.get_top_logprob_indices")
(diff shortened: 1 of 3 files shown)


tests/utils/test_seq2seq_utils.py
```diff
@@ -99,6 +99,55 @@ class FakeOutput:
...
+    # Mock the model
+    mock_model = mock.MagicMock()
+    # Add a fake <pad> token to the generated ids
+    mock_model.generate.return_value = torch.tensor([[0, 1]])
+
+    # Mock the model forward function to return random logits
+    # for a single batch element
+    @dataclass
+    class FakeOutput:
+        logits: torch.tensor
```
elboy3, line 130: you can use a mock for this return value too! 

`mock.MagicMock(logits=torch.rand((1, 1, 20)))`

elboy3 (approved):
one suggestion. sick!

tests/utils/test_seq2seq_utils.py
jonathangomesselman: Oh heck yeah!

bogdan-galileo:
This also takes care of https://app.shortcut.com/galileo/story/7869/dq-logging-breaks-if-a-sample-has-empty-target. Great!

PR #756, opened 2023-10-02
Support both keras 2.13.0 below and above

franz101:
Patch keras until it is fixed here:
https://github.com/tensorflow/tensorflow/pull/62033


setu4993 (approved):
Looks good, nice fix!

bogdan-galileo (approved):
Super nice

PR #757, opened 2023-10-02
Silencing tqdm during generation

jonathangomesselman:

***Description***
This is a very small PR addressing a bug/annoyance when generating in `dq.finish()`. Essentially, during generation we rely no the function `align_tokens_to_character_spans` to create token alignment data. We do this at a per-sample level, since that is how we do generation. However, the `align_tokens_to_character_spans` function was initially developed as a batch processing methods. And with this, we use tqdm.

This becomes annoying when processing at the sample level because we constantly display `tqdm` for just a single iteration loop. See the demonstration bellow.

***Demonstration***

Constantly re-running `tqdm` with a single sample leads to this annoying flickering on the DQ side.

https://github.com/rungalileo/dataquality/assets/29869344/88536639-b9e1-4071-a781-74f147fdf18e

***Solution***

We basically just turn of `tqdm` when generation. Here is the result!

https://github.com/rungalileo/dataquality/assets/29869344/94d36180-7e23-492a-b85e-ac7e8e414209

### DIFF
--- a/dataquality/utils/seq2seq/generation.py
+++ b/dataquality/utils/seq2seq/generation.py
@@ -146,7 +146,9 @@ def generate_on_batch(
         generated_outputs.append(output)
         # Re-tokenize the data to get the token position offsets
         encoded_data = tokenizer([output], return_offsets_mapping=True)
-        aligned_data = align_tokens_to_character_spans(encoded_data["offset_mapping"])
+        aligned_data = align_tokens_to_character_spans(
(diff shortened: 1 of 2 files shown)


jonathangomesselman:
https://app.shortcut.com/galileo/story/7945/dq-silence-tqdm-during-generation

setu4993 (approved):
Nice change.

elboy3 (approved):

PR #759, opened 2023-10-02
Support latest Ultralytics version

franz101:
Support moving from get_settings to SettingsManager and support new imports


tests/integrations/keras/test_end2end_keras.py
```diff
@@ -5,7 +5,12 @@
 import pandas as pd
 import tensorflow as tf
 import vaex
-from keras.engine import data_adapter
+
+try:
+    from keras.engine import data_adapter  # type: ignore
+except ImportError:
+    from tensorflow.python.keras.engine import data_adapter  # type: ignore
```
setu4993, line 12: Isn't this from #756?

pyproject.toml
```diff
@@ -62,7 +62,7 @@ doc = [
     "sphinx-markdown-builder"
 ]
 test = [
-    "ultralytics",
+    "ultralytics<=8.0.162",
```
setu4993, line 65: I'm guessing we aren't really supporting the latest Ultralytics version (PR title) since we're pinning it here and the latest is [v8.0.191](https://pypi.org/project/ultralytics/#history)?

setu4993 (commented):
The Ultralyics changes look fine, but there's some overlap with what should have changed in #756.

There also appears to be a mismatch in how we're supporting this since we're pinning to an old version of `ultralytics`.

elboy3 (changes requested):
Your exception message shows `"This version of dqyolo is only compatible with ultralytics==8.0.190"` but the pyproejct shows >=8.0.190.... let's be consistent

franz101:
> Your exception message shows `"This version of dqyolo is only compatible with ultralytics==8.0.190"` but the pyproejct shows >=8.0.190.... let's be consistent
Changed it to:
"This version of dqyolo is not compatible with ultralytics below 8.0.190."
"Please upgrade ultralytics with `pip install --upgrade ultralytics==8.0.190"
"or higher."

tests/integrations/keras/test_end2end_keras.py
franz101: removed, was causing the tests not to run as m1 only supports keras 2.13.0

pyproject.toml
franz101: updated

elboy3 (approved):

PR #766, opened 2023-10-04
Feature/encoder decoder dq restructure

elboy3:
>
>

***Demonstration***


dataquality/schemas/task_type.py
```diff
@@ -14,7 +14,8 @@ class TaskType(str, Enum):
     object_detection = "object_detection"
     semantic_segmentation = "semantic_segmentation"
     prompt_evaluation = "prompt_evaluation"
-    seq2seq = "seq2seq"
+    seq2seq = "seq2seq"  # TODO Remove
```
elboy3, line 17: we'll need to check with rodrigo if he uses "seq2seq" at all on the UI as well. and we'll have to make some api / runners / rungalileo changes too. 

depending on how much seq2seq is hard coded elsewhere, we should either just rename it to encoder decoder OR use a new task type 10 as encoder decoder and just kinda deprecate 8 seq2seq

dataquality/loggers/model_logger/seq2seq/seq2seq.py
```diff
@@ -58,8 +58,12 @@ def token_map_key(self) -> str:
             return self.inference_name
         return str(self.split)
 
+    @abstractmethod
```
elboy3, line 61: i don't think you need this decorator since it's mainly for the base class (definition in `BaseGalileoModelLogger`)

dataquality/loggers/model_logger/seq2seq/encoder_decoder.py
```diff
@@ -0,0 +1,59 @@
...
+    EncoderDecoderLoggerConfig,
+    encoder_decoder_logger_config,
+)
+from dataquality.loggers.model_logger.seq2seq.seq2seq import Seq2SeqModelLogger
+
+
+class EncoderDecoderModelLogger(Seq2SeqModelLogger):
+    # TODO Add in API so we can use encoder_decoder
+    # __logger_name__ = "encoder_decoder"
+    __logger_name__ = "seq2seq"
```
elboy3, line 15: yup, we'll have to change this to `encoder_decoder`

dataquality/loggers/model_logger/__init__.py
```diff
@@ -1,12 +1,12 @@
 from dataquality.loggers.model_logger import (
     image_classification,
-    seq2seq,
     tabular_classification,
     text_classification,
     text_multi_label,
     text_ner,
 )
 from dataquality.loggers.model_logger.base_model_logger import BaseGalileoModelLogger
+from dataquality.loggers.model_logger.seq2seq import encoder_decoder, seq2seq
```
elboy3, line 9: we should rename the `seq2seq` file (not folder) to `seq2seq_base`

dataquality/loggers/data_logger/seq2seq/seq2seq.py
```diff
@@ -295,32 +284,31 @@ def separate_dataframe(
         return BaseLoggerDataFrames(prob=prob, emb=emb, data=data_df)
 
     @classmethod
+    @abstractmethod
```
elboy3, line 287: i'm pretty sure, the abstract method should only go in the base and should not be implemented, that's how i've always seen it used. lmk if you think otherwise we can search a bit deeper into recommended usage!

dataquality/integrations/seq2seq/hf.py
```diff
@@ -1,18 +1,45 @@
...
+
+    Raises an exception if the user has set / is using an incorrect task_type
+    """
+    if task_type == task_type.seq2seq:  # TODO Change to encoder_decoder
+        return encoder_decoder_logger_config
+
+    # TODO Change to encoder_decoder
+    raise GalileoException(
+        "Galileo's seq2seq watch method is only supported for seq2seq"
+    )
```
elboy3, line 36: i think we can just use the get current task type helpers, since they will have already initialized the project with `dq.init` and we will have the task type stored in the config file

elboy3: Look for other instances of where we do `get_data_logger().logger_config`

elboy3 (commented):
Generally looks great! We'll just need to discuss task type switch and how to best handle that without breaking things, then we're good to go

dataquality/loggers/model_logger/__init__.py
jonathangomesselman: Okay love that idea!

dataquality/loggers/data_logger/seq2seq/seq2seq.py
jonathangomesselman: Pasting in some thoughts from GPT-4

jonathangomesselman: On the topic of can you have partial implementations within an `abstractmethod` here are its thoughts:

Yes, you can provide a partial implementation in an abstract method and still require subclasses to override it. This can be useful when there's some shared logic that you want to include in the base class, but you still want to ensure that subclasses provide their own specific implementations.

Here's how you can do this:

1. Define the abstract method in the base class with the @abstractmethod decorator.
2. Within the abstract method, provide the shared logic.
3. In the subclasses, you can use the super() function to call the base class's implementation before or after adding the subclass-specific logic.
Here's an example to illustrate:
```
from abc import ABC, abstractmethod

class Animal(ABC):

    @abstractmethod
    def speak(self):
        print("The animal makes a sound...")

class Dog(Animal):

    def speak(self):
        super().speak()  # Call the base class's implementation
        print("Woof!")

class Cat(Animal):

    def speak(self):
        super().speak()  # Call the base class's implementation
        print("Meow!")
```
In this example:

- Animal is an abstract base class with an abstract method speak that has a partial implementation.
- Both Dog and Cat subclasses override the speak method. They use super().speak() to call the base class's implementation and then add their own specific logic.

jonathangomesselman: On the topic of should abstractmethod just be in the base-class here were its thoughts:

## GPT-4 talking here ##

It's perfectly acceptable and sometimes even desirable to have abstract methods in subclasses (i.e., in intermediate layers of a class hierarchy). This practice can be useful in various scenarios:

1. Refined Abstraction: Sometimes, the base class is a very general abstraction, and intermediate subclasses refine this abstraction but are still too general to provide a concrete implementation for certain methods. These intermediate classes can introduce new abstract methods that are relevant to their refined abstraction.

2. Hierarchical Design: As you develop a class hierarchy, you may realize that certain behaviors are shared among a subset of subclasses, but not all. In such cases, it's beneficial to introduce an intermediate abstract class that encapsulates these shared behaviors as abstract methods.

3. Flexibility: Introducing abstract methods in intermediate classes provides flexibility. As the design evolves, you can ensure that specific subclasses adhere to certain interfaces without affecting unrelated subclasses.

Here's an example to illustrate:
```
from abc import ABC, abstractmethod

class Vehicle(ABC):
    @abstractmethod
    def move(self):
        pass

class LandVehicle(Vehicle):
    @abstractmethod
    def start_engine(self):
        pass

class Car(LandVehicle):
    def move(self):
        print("The car drives.")

    def start_engine(self):
        print("Car engine started.")

class Boat(Vehicle):
    def move(self):
        print("The boat sails.")
```

In this design, the intermediate abstract class LandVehicle provides a refined abstraction for vehicles that move on land. By introducing the start_engine abstract method, we ensure that any concrete subclass of LandVehicle implements this behavior.

jonathangomesselman: I think that what GPT-4 says makes sense to me. I particularly like the points about **hierarchical design** and **Refined Abstraction**. I think here for example, `calculate_cutoffs` is very specific to just Seq2Seq classes!

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: Okay 👌 yes this seems helpful!

dataquality/integrations/seq2seq/hf.py
```diff
@@ -1,42 +1,57 @@
...
-    task_type = get_task_type()
-    assert task_type == TaskType.seq2seq, "This method is only supported for seq2seq"
     assert isinstance(
         tokenizer, PreTrainedTokenizerFast
     ), "Tokenizer must be an instance of PreTrainedTokenizerFast"
     assert getattr(tokenizer, "is_fast", False), "Tokenizer must be a fast tokenizer"
     for attr in ["encode", "decode", "encode_plus", "padding_side"]:
         assert hasattr(tokenizer, attr), f"Tokenizer must support `{attr}`"
-    seq2seq_logger_config.tokenizer = tokenizer
+    logger_config.tokenizer = tokenizer
```
elboy3, line 49: to get config we could call the get data logger config helper  in this fn

dataquality/integrations/seq2seq/hf.py
```diff
@@ -80,17 +103,19 @@ def watch(
...
+    logger_config = dataquality.get_data_logger().logger_config
+    assert type(logger_config) == EncoderDecoderLoggerConfig
+
     assert isinstance(
         model, PreTrainedModel
     ), "model must be an instance of transformers PreTrainedModel"
     assert model.can_generate(), "model must contain a `generate` method for seq2seq"
 
-    set_tokenizer(tokenizer, max_input_tokens, max_target_tokens)
+    set_tokenizer(tokenizer, logger_config, max_input_tokens, max_target_tokens)
```
elboy3, line 115: see above!

dataquality/loggers/data_logger/seq2seq/encoder_decoder.py
```diff
@@ -0,0 +1,148 @@
...
+    all necessary properties (like `add_eos_token`) are set before setting your
+    tokenizer so as to match the tokenization process to your training process.
+
+    NOTE 2: Unlike EncoderOnly models, EncoderDecoder models explicitly separate the
+    processing of the <Input> and <Target> data. Therefore, we do not need any
+    additional information to isolate / extract information on the <Target> data.
+    """
+
+    # TODO Change to encoder_decoder after updating API
+    __logger_name__ = "seq2seq"  # encoder_decoder
```
elboy3, line 77: agreed it should be encoder_decoder

dataquality/loggers/data_logger/seq2seq/seq2seq_base.py
```diff
@@ -96,7 +83,13 @@ def token_map_key(self) -> str:
             return self.inference_name
         return str(self.split)
 
+    @abstractmethod
```
elboy3, line 86: i think we keep the fn here but just remove the abstractmethod decorator

dataquality/loggers/data_logger/seq2seq/seq2seq_base.py
```diff
@@ -295,32 +277,31 @@ def separate_dataframe(
         return BaseLoggerDataFrames(prob=prob, emb=emb, data=data_df)
 
     @classmethod
+    @abstractmethod
```
elboy3, line 280: same, i think let's just have a base classmethod that does some logic, and if the parent's want to call super() and do extra or want to override they can, but let's not mandate that the parent's **have** to override this fn

dataquality/loggers/logger_config/seq2seq/encoder_decoder.py
```diff
@@ -0,0 +1,10 @@
+from dataquality.loggers.logger_config.seq2seq.seq2seq_base import Seq2SeqLoggerConfig
+
+
+class EncoderDecoderLoggerConfig(Seq2SeqLoggerConfig):
+    # TODO Add comment
+    # This currently is purely a wrapper!
+    pass
```
elboy3, line 7:
```suggestion
    """Logger config for Encoder Decoder
    
    This logger currently has same fields as the base class
    """
```

something like this is fine ^ also you don't need the "pass"

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -58,8 +58,12 @@ def token_map_key(self) -> str:
             return self.inference_name
         return str(self.split)
 
+    @abstractmethod
```
elboy3, line 61: 👋  remove!

dataquality/schemas/task_type.py
```diff
@@ -38,6 +39,7 @@ def get_mapping(task_int: int) -> "TaskType":
             5: TaskType.object_detection,
             6: TaskType.semantic_segmentation,
             7: TaskType.prompt_evaluation,
-            8: TaskType.seq2seq,
+            8: TaskType.seq2seq,  # TODO Remove
+            # 8: TaskType.encoder_decoder,  # TODO add on API side
```
elboy3, line 43: i'm personally now thinking we should actually keep seq2seq and just call it deprecated, and then add 10 as encoder_decoder and 11 as decoder_only ..... it seems too hard to move everything from seq2seq to encoder_decoder across all repos

elboy3 (commented):
generally looking great! let's continue to pair

elboy3:
Closing so @jonathangomesselman can finish cleaning comments and then create his own PR!

PR #767, opened 2023-10-04
feat: auto for seq2seq, alpaca only

elboy3:
## What?

Running seq2seq encoder-decoder is now as easy as

```python
from dataquality.integrations.seq2seq.auto import auto
dataset = "tatsu-lab/alpaca"
auto(dataset)
```

## Note
This PR supports `tatsu-lab/alpaca` dataset only. We will have follow up PRs to support
- simplifying to use Trainer
- extending to use other HF datasets, such as billsum
- supporting local datasets in form of csv (or potentially json)

## Shortcut ticket
https://app.shortcut.com/galileo/story/7604/dq-auto-for-s2s

## Drew inspiration from:
https://colab.research.google.com/drive/1SIk7A85KZqWXoZ_Ri4H6UR-wMDnirPN_?authuser=2#scrollTo=-kyyP8P0W27M


dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+            # TODO: temporary, update
+            ds = ds.select(range(100))
+
+            if "id" not in ds.features:
+                ds = ds.add_column("id", list(range(ds.num_rows)))
+            clean_dd[key] = ds
+        return add_val_data_if_missing(clean_dd, TaskType.seq2seq)
+
+
+def _get_labels(dd: DatasetDict, labels: Optional[List[str]] = None) -> List[str]:
```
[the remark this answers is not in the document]
jonathangomesselman: Or at least we don't need this version of it. This assumes labels are categorical like in `TC`.

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+# The default values in Generation Config
+# Check more out here: https://huggingface.co/docs/transformers/v4.30.0/en/main_classes/text_generation#transformers.GenerationConfig
+MAX_NEW_TOKENS = 16
+TEMPERATURE = 0.2  # Keep this low for now
+TOP_P = 1
+TOP_K = 50
+# Generate params
+GENERATE_ON_TRAIN = False
+# Tokenization params
+MAX_INPUT_LENGTH = 32
```
jonathangomesselman, line 34: We will want to change these haha these are tiny. In general we probably want to default to the model's default values and then allow the user to change

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+MAX_TARGET_LENGTH = 16
+
+# Training params
+LR = 3e-4
```
jonathangomesselman, line 38: We will have to make sure that whenever we do `auto` we get specs for the machine they are using to make sure that these default params can work on their GPU. Otherwise they may get cuda memory errors and then we are caputze.

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+ACCUMULATION_STEPS = 4
+BATCH_SIZE = 4
+
+
+# Taken from the docs of the trainer module:
+# https://github.com/huggingface/transformers/blob/main/examples/pytorch/
+# text-classification/run_glue.py#L434
+def preprocess_function(
```
jonathangomesselman, line 46: Are we using this?

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+                ):
+                    optimizer.step()
+                    optimizer.zero_grad()
+
+                step_loss = loss.detach().cpu().item()
+                train_epoch_loss += step_loss
+
+            train_epoch_loss = train_epoch_loss / len(train_dataloader)
+            train_ppl = torch.exp(torch.Tensor([train_epoch_loss])).float()
+            print(f"{epoch=}: {train_ppl=} {train_epoch_loss=}")
```
jonathangomesselman, line 222: With the Trainer we should ideally get a lot nicer logging!

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+    if test_dataloader:
+        dq.set_epoch_and_split(split=Split.test, epoch=epoch)
+        with torch.no_grad():
+            for step, batch in enumerate(tqdm(test_dataloader)):
+                ids = batch["id"]
+                batch = {k: v.to(device) for k, v in batch.items() if k != "id"}
+                outputs = model(**batch)
+                logits = outputs.logits  # Shape - [bs, bs_seq_ln, vocab]
+                dq.log_model_outputs(logits=logits, ids=ids)
+
```
jonathangomesselman, line 262: We should definitely do some memory clean up here to avoid memory errors during `finish`:
```
del optimizer
del batch
del outputs
torch.cuda.empty_cache()
import gc
gc.collect()
```

dataquality/utils/auto.py
```diff
@@ -48,28 +48,40 @@ def try_load_dataset_dict(
...
     if hf_data:
-        ds = load_dataset(hf_data) if isinstance(hf_data, str) else hf_data
-        assert isinstance(ds, DatasetDict), (
+        dd = load_dataset(hf_data) if isinstance(hf_data, str) else hf_data
+        assert isinstance(dd, DatasetDict), (
             "hf_data must be a path to a huggingface DatasetDict in the hf hub or a "
             "DatasetDict object. If this is just a Dataset, pass it to `train_data`"
         )
-        return ds
+        if demo_format_fns and hf_data in demo_format_fns:
```
jonathangomesselman, line 74: With hf you generally don't need to loop over the data `splits`. The map function will do this implicitly for you.

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    model = T5ForConditionalGeneration.from_pretrained(model_checkpoint)
+
+    # Setup the dataloader
+    data_collator = DataCollatorForSeq2Seq(tokenizer, return_tensors="pt", padding=True)
+
+    dataloaders = {}
+    for key in dd.keys():
+        shuffle = key in ["train", "training"]
+        ds_tokenized = tokenize(
```
jonathangomesselman, line 136: In general you can use the `hf.map` at the dataset dict level and it will handle doing it over the splits. In this situation I guess it does not really make a huge difference since you want to create the dataloader dict as well!

jonathangomesselman (commented):
Added some comments @elboy3 - Looks awesome overall. nothing really major on my end. Some minor params we will want to update and some hugging face comments. Sweet stuff!

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def _log_dataset_dict(
+    dd: DatasetDict, input_col: str = "text", target_col: str = "label"
+) -> None:
+    for key in dd:
+        ds = dd[key]
+        if key in Split.get_valid_keys():
+            if input_col != "text" and "text" in ds.column_names:
+                ds = ds.remove_columns("text")
```
bogdan-galileo, line 112: alternatively I would relabel to "text_old" or something. Maybe they want it as metadata for example

dataquality/utils/auto.py
```diff
@@ -92,12 +104,15 @@ def add_val_data_if_missing(dd: DatasetDict) -> DatasetDict:
...
+            label_col = col
             break
-    assert label_col in ds_train.features, "Must have label, ner_tags, or tags"
-    # Can only stratify by a ClassLabel
-    is_classlabel = isinstance(ds_train.features[label_col], ClassLabel)
+    is_classlabel = False
+    if task_type and task_type != TaskType.seq2seq:
+        assert label_col in ds_train.features, "Must have label, ner_tags, or tags"
+        # Can only stratify by a ClassLabel
+        is_classlabel = isinstance(ds_train.features[label_col], ClassLabel)
```
bogdan-galileo, line 115: Seems like this is not backward compatible, no ? The code in the if will never be reached unless we change the way it's called in NER, TC, etc by passing the task type

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+)
+
+import dataquality as dq
+from dataquality.integrations.seq2seq.hf import watch
+from dataquality.schemas.split import Split
+
+# Generation params
+# The default values in Generation Config
+# Check more out here: https://huggingface.co/docs/transformers/v4.30.0/en/main_classes/text_generation#transformers.GenerationConfig
+MAX_NEW_TOKENS = 16
```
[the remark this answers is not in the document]
bogdan-galileo: Great point. I feel like a bunch of these are
- dataset specific: max_input_length and max_output_length for example
- use-case specific: max_new_tokens, temperature, etc
but we can't have them all as args in `auto`.

Putting them in a class would also allow changing them easier by
- user instantiates that config class
- changes whatever values
- passes that object to dq.auto!

Simply Magic 🪄

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    manager = S2SDatasetManager()
+    dd = manager.get_dataset_dict(
+        hf_data,
+        train_data=train_data,
+        val_data=val_data,
+        test_data=test_data,
+    )
+    dq.login()
+    a.log_function("auto/tc")
```
bogdan-galileo, line 240: auto/seq2seq

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+        if key in Split.get_valid_keys():
+            if input_col != "text" and "text" in ds.column_names:
+                ds = ds.remove_columns("text")
+            if target_col != "label" and "label" in ds.column_names:
+                ds = ds.remove_columns("labels")
+
+            dq.log_dataset(ds, text=input_col, label=target_col, split=key)
+
+
+def auto(
```
bogdan-galileo, line 119: We'll need a mechanism for them to be able to pass input_colname and target_colname, etc (and a bunch of other things). In LLMs the colnames are so wild between summary, article, document, text, etc, we won't be able to guess it and we can't ask them to rename their dataset.

Maybe we could have a few configs
- DatasetConfig containing input_colname, max_input_tokens, etc
- TrainingConfig containing epoch, model_name, etc
- GenerationConfig
as args, that would clean up the list of args!

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+    wait: bool,
+    create_data_embs: Optional[bool] = None,
+) -> Trainer:
+    # training and evaluation
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    model = model.to(device)
+
+    train_dataloader = dataloaders.get(Split.training) or dataloaders.get(Split.train)
+    eval_dataloader = dataloaders.get(Split.validation) or dataloaders.get(Split.val)
+    test_dataloader = dataloaders.get(Split.test)
```
bogdan-galileo, line 189: There is also `valid` and `testing`

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+            eval_epoch_loss = eval_epoch_loss / len(eval_dataloader)
+            eval_ppl = torch.exp(torch.Tensor([eval_epoch_loss])).item()
+
+        print(
+            f"{epoch=}: {train_ppl=} {train_epoch_loss=} {eval_ppl=} {eval_epoch_loss=}"
+        )
+
+    # After training, do test step
+    if test_dataloader:
+        dq.set_epoch_and_split(split=Split.test, epoch=epoch)
```
bogdan-galileo, line 254: should put the model in eval mode `model.eval()`

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+    train_dataloader = dataloaders.get(Split.training) or dataloaders.get(Split.train)
+    eval_dataloader = dataloaders.get(Split.validation) or dataloaders.get(Split.val)
+    test_dataloader = dataloaders.get(Split.test)
+
+    optimizer = Adafactor(
+        model.parameters(), lr=LR, scale_parameter=False, relative_step=False
+    )
+
+    for epoch in range(num_epochs):
+        if train_dataloader:
```
bogdan-galileo, line 196: TODO: we should train with mixed precision to make it faster. Can make a follow-up ticket to bake that in

dataquality/integrations/seq2seq/s2s_trainer.py
franz101: good point, you can do just datasets.map(tokenizer)

instead of
for ds in datasets:
     ds.map...

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+from dataquality.dq_auto.base_data_manager import BaseDatasetManager
+from dataquality.integrations.seq2seq.s2s_trainer import do_train, get_trainer
+from dataquality.schemas.split import Split
+from dataquality.schemas.task_type import TaskType
+from dataquality.utils.auto import (
+    add_val_data_if_missing,
+    run_name_from_hf_dataset,
+)
+
+a = Analytics(ApiClient, dq.config)
```
setu4993, line 19: I don't think we need this anymore?

dataquality/utils/auto.py
```diff
@@ -48,28 +48,40 @@ def try_load_dataset_dict(
...
+    and `target` columns instead of `text` and `label`
     """
     if all([hf_data is None, train_data is None]):
         hf_data = choice(demo_datasets)
         print(f"No dataset provided, using {hf_data} for run")
     if hf_data:
-        ds = load_dataset(hf_data) if isinstance(hf_data, str) else hf_data
-        assert isinstance(ds, DatasetDict), (
+        dd = load_dataset(hf_data) if isinstance(hf_data, str) else hf_data
+        assert isinstance(dd, DatasetDict), (
```
setu4993, line 70: Good change.

dataquality/utils/auto.py
setu4993: Yeah, we should reuse that nicety.

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+
+class S2SDatasetManager(BaseDatasetManager):
+    DEMO_DATASETS = [
+        "tatsu-lab/alpaca",
+        # "billsum",
+    ]
+
+    DATASET_FORMAT_FNS = {
+        "tatsu-lab/alpaca": format_alpaca,
+    }
```
setu4993, line 56: It feels super weird to have a map here instead of a single fn but I get _why_ we have that.

It feels limiting because we're matching the exact key of the dataset, which we're unlikely to match in most cases.

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+
+    DATASET_COLS = {
+        "tatsu-lab/alpaca": {
+            "input": "formatted_input",
+            "target": "output",
+        }
+    }
```
setu4993, line 63: Given we have multiple such properties, feels like we could use a different abstraction layer than one within here.

dataquality/integrations/seq2seq/hf.py
```diff
@@ -66,7 +66,7 @@ def watch(
     model: PreTrainedModel,
     tokenizer: PreTrainedTokenizerFast,
     generation_config: GenerationConfig,
-    generatation_splits: Optional[List[str]] = None,
+    generation_splits: Optional[List[str]] = None,
```
setu4993, line 69: :)

dataquality/utils/auto.py
setu4993: Well, it is backward compatible because it defaults to `None`, but yeah, the code here won't be reached, which means we won't get the validations we are expecting from those task types.

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+    if hasattr(train_labels, "names"):
+        return train_labels.names
+    return sorted(set(dd[Split.train]["label"]))
+
+
+def _log_dataset_dict(
+    dd: DatasetDict, input_col: str = "text", target_col: str = "label"
+) -> None:
+    for key in dd:
+        ds = dd[key]
```
setu4993, line 109: I forget what HF datasets' structure is, but I'm almost sure you could do:

```suggestion
    for key, ds in dd.items():
```

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+
+        auto()
+    ```
+
+    An example using `auto` with a hosted huggingface dataset
+    ```python
+        from dataquality.auto.text_classification import auto
+
+        auto(hf_data="rungalileo/trec6")
+    ```
```
setu4993, line 192: Entire docstring could use a read-through + fixes.

dataquality/integrations/seq2seq/s2s_trainer.py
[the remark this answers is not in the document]
setu4993: Agreed!

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+
+# Training params
+LR = 3e-4
+ACCUMULATION_STEPS = 4
+BATCH_SIZE = 4
+
+
+# Taken from the docs of the trainer module:
+# https://github.com/huggingface/transformers/blob/main/examples/pytorch/
+# text-classification/run_glue.py#L434
```
setu4993, line 45: TC example?

setu4993 (commented):
Looks like a good v1. Left a few comments, but I'll delegate to @bogdan-galileo and @jonathangomesselman for approvals.

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+    hf_model: str = "google/flan-t5-base",
+    project_name: str = "auto_s2s",
+    run_name: Optional[str] = None,
+    wait: bool = True,
+    max_input_tokens: Optional[int] = None,
+    max_target_tokens: Optional[int] = None,
+    create_data_embs: Optional[bool] = None,
+    generation_splits: Optional[List[str]] = None,
+) -> Trainer:
+    """Automatically gets insights on a text classification dataset
```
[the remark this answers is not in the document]
elboy3: definitely

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,263 @@
...
+        val_data=val_data,
+        test_data=test_data,
+    )
+    dq.login()
+    a.log_function("auto/tc")
+    if not run_name and isinstance(hf_data, str):
+        run_name = run_name_from_hf_dataset(hf_data)
+
+    dq.init(TaskType.seq2seq, project_name=project_name, run_name=run_name)
+    input_col = "text"
```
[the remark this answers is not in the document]
elboy3: yes! will add support for custom local datasets next via args for the rename, or some kind of `DatasetConfig` like Bogdan suggested

dataquality/integrations/seq2seq/s2s_trainer.py
elboy3: Oh yeah, okay I'll clean this up!

dataquality/integrations/seq2seq/s2s_trainer.py
elboy3: Yeah good call out. I changed it elsewhere to run on the whole datasetdict, but with DL creation maybe we keep in for loop for now? do y'all have a strong opinion here?

bogdan-galileo (approved):
My approval won't unblock merging but LGTM

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,187 @@
...
+        hf_model,
+        input_col,
+        target_col,
+        max_input_tokens,
+        max_target_tokens,
+        generation_splits,
+    )
+
+    _log_dataset_dict(dd, input_col=input_col, target_col=target_col)
+    return do_train(model, dataloaders, num_train_epochs, wait, create_data_embs)
```
dcaustin33, line 187: We should warn in here training will take forever if they do not have a cuda device

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,254 @@
...
+        model_checkpoint, use_fast=True, model_max_length=max_input_tokens
+    )
+    model = T5ForConditionalGeneration.from_pretrained(model_checkpoint)
+
+    # Setup the dataloader
+    data_collator = DataCollatorForSeq2Seq(tokenizer, return_tensors="pt", padding=True)
+
+    dataloaders = {}
+    for key in dd.keys():
+        shuffle = key in ["train", "training"]
```
dcaustin33, line 122: Either way works but there is no harm to just setting shuffle to True all the time

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -0,0 +1,254 @@
...
+) -> Trainer:
+    # training and evaluation
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    model = model.to(device)
+
+    train_dataloader = dataloaders.get(Split.training) or dataloaders.get(Split.train)
+    eval_dataloader = dataloaders.get(Split.validation) or dataloaders.get(Split.val)
+    test_dataloader = dataloaders.get(Split.test)
+
+    optimizer = Adafactor(
```
dcaustin33, line 178: Lets be geniuses and use 8 bit optimizer from bitsandbytes will save a lot of mem

jonathangomesselman: Hahaha how stable are these? Idk if I was just doing things wrong but when I've used Pytorch mixed precision I had issues with exploding activations / Nans

dataquality/integrations/seq2seq/formatter.py
```diff
@@ -0,0 +1,56 @@
+from dataclasses import dataclass
+from typing import Dict, Type
+
+
+@dataclass
+class BaseFormatter:
+    name: str = ""
```
setu4993, line 7: This should be a required property.

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -179,34 +179,36 @@ def do_train(
...
+            ):
+                optimizer.step()
+                optimizer.zero_grad()
+
+            step_loss = loss.detach().cpu().item()
+            train_epoch_loss += step_loss
+
+        train_epoch_loss = train_epoch_loss / len(train_dataloader)
+        train_ppl = torch.exp(torch.Tensor([train_epoch_loss])).float()
+        print(f"{epoch=}: {train_ppl=} {train_epoch_loss=}")
```
setu4993, line 211: Do we really want to print this?

dataquality/dq_auto/base_data_manager.py
```diff
@@ -1,20 +1,23 @@
...
-    try_load_dataset_dict,
 )
 
 
 class BaseDatasetManager:
     DEMO_DATASETS: List[str] = []
-    DATASET_FORMAT_FNS: Dict[str, Callable] = {}
+
+    def __init__(self) -> None:
+        self.formatter = BaseFormatter()
```
setu4993, line 20: Should this be initialized to `None`?

dataquality/integrations/seq2seq/auto.py
```diff
@@ -54,12 +54,13 @@ def _validate_dataset_dict(
...
-    for key, ds in dd.items():
+    for key in dd.keys():
+        ds: Dataset = dd[key]
         if key in Split.get_valid_keys():
             if input_col != "text" and "text" in ds.column_names:
-                ds = ds.remove_columns("text")
+                ds = ds.rename_columns({"text": "_metadata_text"})
             if target_col != "label" and "label" in ds.column_names:
-                ds = ds.remove_columns("labels")
+                ds = ds.rename_columns({"label": "_metadata_label"})
```
setu4993: FWIW, this should also work on the dataset dict without having to iterate through it.

setu4993 (commented):
Looking good! There's a bunch of new stuff here, so would love to see some tests in here before we merge.

dataquality/integrations/seq2seq/auto.py
elboy3: @setu4993 that's true, but if the col is in one of the datasets and not another it will raise an error. do you know a workaround for that?

dataquality/integrations/seq2seq/s2s_trainer.py
elboy3: for now yes 🤓  we can remove in future

dataquality/integrations/seq2seq/auto.py
```diff
@@ -0,0 +1,187 @@
...
+from dataquality.utils.auto import (
+    add_val_data_if_missing,
+    run_name_from_hf_dataset,
+)
+
+
+class S2SDatasetManager(BaseDatasetManager):
+    DEMO_DATASETS = [
+        "tatsu-lab/alpaca",
+        # "billsum",
```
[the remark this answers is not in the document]
elboy3: keep for now, as a todo extension!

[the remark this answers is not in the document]
elboy3: will add a todo in comment

dataquality/integrations/seq2seq/auto.py
setu4993: We could just have a fn and apply that fn!

dataquality/dq_auto/base_data_manager.py
setu4993: Ah, if you want the default values then this is fine.

dataquality/integrations/seq2seq/formatter.py
```diff
@@ -54,4 +67,4 @@ def get_formatter(name: str) -> BaseFormatter:
 
     If the name isn't found, returns the base formatter
     """
-    return FORMATTER_MAPPING.get(name, BaseFormatter)()
+    return FORMATTER_MAPPING.get(name, DefaultFormatter)()  # type: ignore
```
setu4993, line 70: Can we add a note here about why we're ignoring the type?

setu4993 (approved):
Looks neat, excited to be able to use this!

PR #769, opened 2023-10-10
Free GPU Memory Before `dq.finish()` in `auto()`

jonathangomesselman:

This PR helps prevent potential `CUDA` memory errors in `dq.auto()`. After training, `dq.finish()` uses some additional `GPU` memory for computing embeddings and for generation. If after training, there is very little available GPU RAM, we can run into big trouble!

Thankfully though, after training is complete, a lot of memory / data on the GPU is no longer needed for `dq.finish()`. For example, a big memory "hog" is the `optimizer`, which for many optimizers, effectively stores multiple copies of the entire model weights. 

To reclaim / free unused GPU memory, and pretty much safely ensure that `dq.finish()` will not error out we use the following pattern:
```
del  optimizer
...
torch.cuda.empty_cache()
import gc
gc.collect()
```

***Demonstration***

To give a bit of additional context into how much this actually saves, here are some rough numbers:

I was training a `flan-t5-small` model with `batch_size=8`. At the end of training / during training I was pretty consistently using around `7.7GB` of GPU RAM.

After freeing unused variables e.g. `optimizer, batch, and outputs`, we went down to `~3.5GB` of GPU RAM. Massive savings!!

To round out the example, when calling `dq.finish()` the GPU RAM usage goes back up to `~4.4GB`. 

So overall, we see that freeing unused variables effectively makes up for all of the additional memory used during `dq.finish()`. Moreover, we emphasize how important this is, since even the small model uses `7.7GB` of memory during training and an additional `<1GB` during `.finis()`! Often people will push their training right to the memory limit (testing out different batch sizes to find the sweet spot), so even if `dq.finish()` only uses `<1GB` memory, we can definitely run into issues.

### DIFF
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -251,6 +251,6 @@ def do_train(
                 logits = outputs.logits  # Shape - [bs, bs_seq_ln, vocab]
                 dq.log_model_outputs(logits=logits, ids=ids)
 
-    cleanup_cuda(optimizer, batch, outputs)
+    cleanup_cuda(optimizer, [batch, outputs])
     dq.finish(wait=wait, create_data_embs=create_data_embs)
     return model
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -1,3 +1,4 @@
+import gc
 from typing import Dict, List, Optional, Tuple
 
 import torch
@@ -249,6 +250,14 @@ def do_train(
                 outputs = model(**batch)
                 logits = outputs.logits  # Shape - [bs, bs_seq_ln, vocab]
                 dq.log_model_outputs(logits=logits, ids=ids)
+
+    # Delete unused variables to free CUDA memory to ensure that
+    # dq.finish() does not run into out-of-memory errors.
+    del optimizer  # This is the most important!
+    del batch
+    del outputs
+    torch.cuda.empty_cache()
+    gc.collect()
 
     dq.finish(wait=wait, create_data_embs=create_data_embs)
     return model
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -253,6 +253,6 @@ def do_train(
 
     # Cleanup all unused data on the GPU and any references
     # to that data
-    cleanup_cuda(optimizer, [batch, outputs])
+    cleanup_cuda(optimizer, [logits, loss, batch, outputs])
     dq.finish(wait=wait, create_data_embs=create_data_embs)
     return model
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -1,4 +1,3 @@
-import gc
 from typing import Dict, List, Optional, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


jonathangomesselman:
https://app.shortcut.com/galileo/story/7806/dq-gpu-memory-improvement-for-auto

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -250,5 +251,13 @@ def do_train(
                 logits = outputs.logits  # Shape - [bs, bs_seq_ln, vocab]
                 dq.log_model_outputs(logits=logits, ids=ids)
 
+    # Delete unused variables to free CUDA memory to ensure that
+    # dq.finish() does not run into out-of-memory errors.
+    del optimizer  # This is the most important!
+    del batch
+    del outputs
+    torch.cuda.empty_cache()
+    gc.collect()
```
elboy3, line 260: can we put this in a `cleanup()` function? 

otherwise looks great.

elboy3 (approved):

setu4993 (approved):
Nice improvement. A good change could be for us to switch over to a better where auto isn't a single fn but a class, where cleanup would automatically occur once training is complete.

Something to think about for the evolution of dq, @elboy3 and @franz101.

dataquality/utils/torch.py
```diff
@@ -27,6 +27,19 @@
...
+    """Cleanup cuda memory
+
+    Delete unused variables to free CUDA memory to ensure that
+    dq.finish() does not run into out-of-memory errors.
+    """
+    for arg in args:
+        del arg
+
+    torch.cuda.empty_cache()
+    gc.collect()
```
bogdan-galileo, line 40: I don't know if t matters, but I'm usually seeing this in the other order, first gc.collect and then torch.cuda.empty_cache. It also makes sense to me to first garbage collect and once it's collected we can clear up its allocation from GPU too.

jonathangomesselman: Interesting! I honestly took this from your notebook where you had it the other way?

dataquality/integrations/seq2seq/s2s_trainer.py
jonathangomesselman: Sure let me give that a go!

dataquality/utils/torch.py
elboy3: From GPT:

```
The order in which you use gc.collect() and torch.cuda.empty_cache() can depend on your specific use case, but generally, it's a good practice to use gc.collect() after torch.cuda.empty_cache() for the following reasons:

torch.cuda.empty_cache() releases GPU memory occupied by PyTorch tensors.

gc.collect() is the Python garbage collection mechanism that collects and frees up CPU memory used by Python objects that are no longer referenced.

By emptying the GPU memory with torch.cuda.empty_cache() first, you release GPU memory explicitly. Then, running gc.collect() can help clean up any residual CPU memory that might have been used in the process of clearing GPU memory.
```

jonathangomesselman: Ohh I don't think that is how you want to check for memory used!

jonathangomesselman: I have seen this referenced in the `PyTorch` community:
```
def memory_stats():
    print(torch.cuda.memory_allocated()/1024**2)
    print(torch.cuda.memory_cached()/1024**2)
```

bogdan-galileo: Maybe the tensors are too small and the fluctuation of RAM is bigger than just cleaning them up ? Maybe using 10^6 instead of 10^3 ?

elboy3: yeah... i think we just merge cause we know it works and we come back to this test later. 

it is weird cause these tests aren't running on cuda

PR #773, opened 2023-10-12
fix: s2s auto small improvements

elboy3:
- increment `max_new_tokens` to 64
- Remove train print statements after eval loop
- Optionally return model (default to false, delete memory)

### DIFF
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -232,6 +232,6 @@ def do_train(
 
     # Cleanup all unused data on the GPU and any references
     # to that data
-    cleanup_cuda(optimizer, [logits, loss, batch, outputs])
+    cleanup_cuda(optimizer=optimizer, tensors=[logits, loss, batch, outputs])
     dq.finish(wait=wait, create_data_embs=training_config.create_data_embs)
     return model
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -14,7 +14,6 @@ from transformers import (
     PreTrainedModel,
     PreTrainedTokenizerFast,
     T5ForConditionalGeneration,
-    Trainer,
 )
 
 import dataquality as dq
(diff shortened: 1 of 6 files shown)


jonathangomesselman (approved):
🚀 🚢

dataquality/integrations/seq2seq/schema.py
```diff
@@ -71,7 +71,7 @@ class Seq2SeqGenerationConfig:
         default to ["test"]
     """
 
-    max_new_tokens: int = 16
+    max_new_tokens: int = 64
```
jonathangomesselman, line 74: One day we will change this haha

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -217,9 +216,7 @@ def do_train(
             eval_epoch_loss = eval_epoch_loss / len(eval_dataloader)
             eval_ppl = torch.exp(torch.Tensor([eval_epoch_loss])).item()
 
-        print(
-            f"{epoch=}: {train_ppl=} {train_epoch_loss=} {eval_ppl=} {eval_epoch_loss=}"
-        )
+        print(f"{epoch=}: {eval_ppl=} {eval_epoch_loss=}")
```
dcaustin33, line 219: swanky

dataquality/integrations/seq2seq/schema.py
dcaustin33, line 74: If this is for the generation step I think it is worth thinking about "smartly" setting it with regards to the dataset

dataquality/utils/torch.py
```diff
@@ -28,12 +29,20 @@
...
+    tensors: Optional[List[torch.Tensor]] = None,
+) -> None:
     """Cleanup cuda memory
 
     Delete unused variables to free CUDA memory to ensure that
     dq.finish() does not run into out-of-memory errors.
     """
+    tensors = tensors or []
+    if model:
+        del model
```
dcaustin33, line 44: I think transfer to cpu mem first and then delete to avoid any CUDA oddities. I have had in the past where I delete stuff on cuda and it doesn't get collected for some reason but transferring to CPU always seems to clear that mem block.

dcaustin33 (approved):
Looks good! Only adjustment I would make is transferring to cpu before deleting model

dataquality/integrations/seq2seq/schema.py
jonathangomesselman: Yeah I think we should likely have it just match `max_target_tokens`. Since this is the max tokens we train over

dataquality/utils/torch.py
jonathangomesselman: Hmm interesting! This was partly why we used `torch.cuda.clear_cache()`. But cool if this essentially does the same thing?

PR #777, opened 2023-10-17
feat: s2s auto metadata

elboy3:
https://app.shortcut.com/galileo/story/8226/dq-auto-support-for-logging-metadata

Tested locally and it works

https://console.dev.rungalileo.io/insights/4194b985-6c2c-459b-9199-5ed53135ce5a/dd896833-bf7f-4c2a-b51e-6aa8ede6ccaa?dataframeColumns=Unnamed%3A+0%3B%26category%3B%26data_error_potential%3B%26input%3B%26lang%3B%26perplexity%3B%26raw_data_hash%3B%26source%3B%26target&taskType=8

### DIFF
--- a/dataquality/loggers/data_logger/seq2seq.py
+++ b/dataquality/loggers/data_logger/seq2seq.py
@@ -124,16 +124,17 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
         self.logger_config.id_to_tokens[self.token_map_key].update(id_to_tokens)
 
     def _get_input_df(self) -> DataFrame:
-        inp = {
-            C.id.value: self.ids,
-            C.text.value: self.texts,
-            C.label.value: self.labels,
-            C.split_.value: [self.split] * len(self.ids),
-            C.token_label_positions.value: pa.array(self.token_label_positions),
-            C.token_label_offsets.value: pa.array(self.token_label_offsets),
-            **self.meta,
-        }
-        return vaex.from_dict(inp)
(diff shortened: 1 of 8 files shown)


dataquality/loggers/data_logger/seq2seq.py
```diff
@@ -124,16 +124,16 @@ def validate_and_format(self) -> None:
...
-            }
-        )
+        inp = {
+            C.id.value: self.ids,
+            C.text.value: self.texts,
+            C.label.value: self.labels,
+            C.split_.value: [self.split] * len(self.ids),
+            C.token_label_positions.value: pa.array(self.token_label_positions),
+            C.token_label_offsets.value: pa.array(self.token_label_offsets),
+            **self.meta,
```
setu4993, line 134: Just for my understanding: This could have been as a part of the previous way it was done, too, right? Or is there something special / specific here for which it needs to be different?

setu4993 (approved):
Would be nice to add a couple tests for this.

dataquality/loggers/data_logger/seq2seq.py
elboy3: ahh yes it can, i'll fix. i broke it into 2 separate cause i had a breakpoint in there, but will revert

setu4993: No worries, this is fine, I was just asking because it was a larger diff than it should've been :).

PR #778, opened 2023-10-19
fix: auto alpaca

elboy3:
We were running into some column renaming issues that I fix in this PR

I also made some edits to allow the user to pass in an upper limit to the dataset size that is configurable in DatasetConfig

A HF dataset has a default size limit that can be upped by the user

### DIFF
--- a/dataquality/dq_auto/schema.py
+++ b/dataquality/dq_auto/schema.py
@@ -4,6 +4,8 @@ from typing import Optional, Union
 import pandas as pd
 from datasets import Dataset, DatasetDict
 
+from dataquality.integrations.seq2seq.formatter import BaseFormatter, DefaultFormatter
+
 
 @dataclass
 class BaseAutoDatasetConfig:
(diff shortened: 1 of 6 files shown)


setu4993 (changes requested):
Creds in notebook again...

dataquality/dq_auto/schema.py
```diff
@@ -46,6 +48,9 @@ class BaseAutoDatasetConfig:
     # Column names
     input_col: str = "text"
     target_col: str = "label"
+    # Dataset input / output formatter
+    max_train_size: Optional[int] = None
+    formatter: BaseFormatter = DefaultFormatter()
```
setu4993, line 53: This is a more idiomatic way of initializing for a dataclass:

```suggestion
from dataclasses import field
...
    formatter: BaseFormatter = field(default_factory=DefaultFormatter)
```

setu4993 (commented):
This looks good. Left a suggestion.

Let's clean up the notebook and creds before merging.

setu4993 (approved):

PR #779, opened 2023-10-23
feat: s2s auto chat support

elboy3:
V1 of auto chat support. This just unravels the turns and has input as user, target as assistant

V2 will include some of the chat history

https://app.shortcut.com/galileo/story/8388/dq-support-basic-chat-models

Did e2e tests for:
✅ chat data [link](https://console.dev.rungalileo.io/insights/a29cfac8-773d-45f2-aaac-adfd61fe6ef5/c3399705-5935-445f-ab18-0917349af121?dataframeColumns=Chatbot_skip_use_as_completion%3B%26Chatbot_toxicity%3B%26User_skip_use_as_completion%3B%26User_toxicity%3B%26data_error_potential%3B%26input%3B%26perplexity%3B%26raw_data_hash%3B%26target&taskType=8)
✅ auto with alpaca [link](https://console.dev.rungalileo.io/insights/a29cfac8-773d-45f2-aaac-adfd61fe6ef5/3f02d3f4-0f9b-4d8a-a634-9276e1d14174?insightsTab=metrics&taskType=8)
✅ completion dataset

PR #780, opened 2023-10-24
fix: history chat

elboy3:
Follow up of https://github.com/rungalileo/dataquality/pull/779 to support having history in the chat of previous turns

Builds off of Jon's work here: https://colab.research.google.com/drive/1O5v41k-sMTBxTVqQCAkDuA8q5RX3qLZL?authuser=2#scrollTo=RhqLFpEPq2sI

Demo run: https://console.dev.rungalileo.io/insights/a29cfac8-773d-45f2-aaac-adfd61fe6ef5/201fc929-63be-4341-82e6-a694900719dc?dataframeColumns=chat_id%3B%26data_error_potential%3B%26input%3B%26perplexity%3B%26project_name%3B%26target%3B%26turn_id&metaFilter=chat_id%3A5&sortBy=turn_id&sortDirection=asc&taskType=8

### DIFF
--- a/dataquality/integrations/seq2seq/formatters/chat.py
+++ b/dataquality/integrations/seq2seq/formatters/chat.py
@@ -96,3 +96,50 @@ class ChatFormatter(BaseFormatter):
                 raise ValueError(f"Role {role} not recognized")
 
         return unraveled_turns
+
+
+@dataclass
+class ChatHistoryFormatter(ChatFormatter):
+    def format_sample(
+        self, sample: Dict[str, Any], idx: Optional[int] = None
+    ) -> Dict[str, Any]:
+        """Formats a chat dataset for seq2seq with previous turn history
+
+        Similair to ChatFormatter, except subsequent turns contain
+        context from previous turns.
+
+        Example:
+            >>> sample = {
+            ...     "turns": [
+            ...         {"role": "User", "content": "Hello"},
+            ...         {"role": "Chatbot", "content": "Hi"},
+            ...         {"role": "User", "content": "How are you?"},
+            ...         {"role": "Chatbot", "content": "I'm good, how are you?"},
+            ...     ],
+            ...     "metadata": {"unique_id": 1234, "dataset": "test"},
+            ...     "score": 0.5,
+            ... }
+            >>> ChatFormatter().format_sample(sample, 5)
+            {
+                "chat_id": [5, 5],
+                "turn_id": [1, 2],
+                "input": ["Hello", "Hello\n\nHi\n\nHow are you?"],
+                "target": ["Hi", "I'm good, how are you?"],
+                "unique_id": [1234, 1234],
+                "dataset": ["test", "test"],
+            }
+        """
+        formatted_sample = super().format_sample(sample, idx)
+        # Add previous turns to input
+        user_col = formatted_sample[self.input_col]
+        assistant_col = formatted_sample[self.target_col]
+        for i in range(len(user_col)):
+            history_input = ""
+            if i > 0:
+                history_input += f"{user_col[i - 1]}\n\n"
+                history_input += f"{self.assistant}: {assistant_col[i - 1]}\n\n"
+
+            history_input += f"{self.user}: {user_col[i]}"
+            formatted_sample[self.input_col][i] = history_input
+
+        return formatted_sample
--- a/dataquality/integrations/seq2seq/formatters/chat.py
+++ b/dataquality/integrations/seq2seq/formatters/chat.py
@@ -184,6 +184,12 @@ class ChatHistoryFormatter(ChatFormatter):
             parsed_history = self.tokenizer.decode(
                 history_tokens, skip_special_tokens=True
             )
-            user_inputs[i] = parsed_history.replace(NEWLINE_TOKEN, NEWLINE_CHAR)
+            parsed_history = parsed_history.replace(NEWLINE_TOKEN, NEWLINE_CHAR)
+            # We don't want to have the input start mid turn, so we find the first
+            # instance of the user or assistant role and start there
+            first_user_index = parsed_history.find(f"{self.user}: ")
+            first_assistant_index = parsed_history.find(f"{self.assistant}: ")
+            start_index = min(first_user_index, first_assistant_index)
+            user_inputs[i] = parsed_history[start_index:]
 
         return formatted_sample
--- a/dataquality/integrations/seq2seq/formatters/chat.py
+++ b/dataquality/integrations/seq2seq/formatters/chat.py
@@ -6,6 +6,11 @@ from transformers import AutoTokenizer, PreTrainedTokenizerFast
 
 from dataquality.integrations.seq2seq.formatters.base import BaseFormatter
 
+# HF tokenizers don't support newlines, so we use a token to represent them
+# Example of tokenizer without the NEWLINE token:
+# result = tokenizer.decode(tokenizer("Hello\n\n\nWorld"))
+# print(result)
+# >>> HelloWorld
 NEWLINE_CHAR = "\n\n\n"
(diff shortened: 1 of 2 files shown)


dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -96,3 +96,50 @@ def format_sample(
...
+        # Add previous turns to input
+        user_col = formatted_sample[self.input_col]
+        assistant_col = formatted_sample[self.target_col]
+        for i in range(len(user_col)):
+            history_input = ""
+            if i > 0:
+                history_input += f"{user_col[i - 1]}\n\n"
+                history_input += f"{self.assistant}: {assistant_col[i - 1]}\n\n"
+
+            history_input += f"{self.user}: {user_col[i]}"
```
elboy3, line 142: my thought is we could tokenize and just take the last 512 (or however many set in the config) tokens. @jonathangomesselman would that work?

jonathangomesselman: I think I am a bit confused: Shouldn't `history_input` be outside this for loop? Otherwise we are just reseting the history each loop?

jonathangomesselman: This could be a great question to discuss with customers! I am not sure exactly is the better approach honestly.

dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -96,3 +101,79 @@ def format_sample(
...
+                model_max_length=self.max_input_tokens,
+            )
+            self.tokenizer.add_tokens([NEWLINE_TOKEN])
+
+    def format_sample(
+        self, sample: Dict[str, Any], idx: Optional[int] = None
+    ) -> Dict[str, Any]:
+        """Formats a chat dataset for seq2seq with previous turn history
+
+        Similair to ChatFormatter, except subsequent turns contain
```
elboy3, line 127:
```suggestion
        Similar to ChatFormatter, except subsequent turns contain
```

dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -96,3 +101,79 @@ def format_sample(
...
+
+
+@dataclass
+class ChatHistoryFormatter(ChatFormatter):
+    hf_tokenizer: Optional[str] = "google/flan-t5-base"
+    tokenizer: Optional[PreTrainedTokenizerFast] = None
+    max_input_tokens: int = 512
+
+    def __post_init__(self) -> None:
+        if self.hf_tokenizer:
```
setu4993, line 113: FWIW, this would _always_ check to True since the default value is an str. We'd need to explicitly be passed in a `None` for this to not set.

dataquality/integrations/seq2seq/formatters/chat.py
setu4993: There's a few different ways this could be done:
1. Take the last 512, irrespective of turns.
2. Greedily consume the most recent turns until we get as close to 512.

I think for v1 maybe we should just do 2 and then take it to customers so that maybe they can customize it.

setu4993 (commented):
Left a few questions / notes. Let me know when you want another review!

dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -96,3 +101,79 @@ def format_sample(
...
+            self.tokenizer = AutoTokenizer.from_pretrained(
+                self.hf_tokenizer,
+                truncation_side="left",
+                use_fast=True,
+                model_max_length=self.max_input_tokens,
+            )
+            self.tokenizer.add_tokens([NEWLINE_TOKEN])
```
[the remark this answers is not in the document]
elboy3: good call

dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -96,3 +101,79 @@ def format_sample(
...
+                )
+
+            history_input += f"{self.user}: {user_col[i]}"
+            history_input = history_input.replace(NEWLINE_CHAR, NEWLINE_TOKEN)
+            # Only take the last max_input_tokens
+            history_tokens = self.tokenizer(history_input, truncation=True)["input_ids"]
+            parsed_history = self.tokenizer.decode(
+                history_tokens, skip_special_tokens=True
+            )
+            formatted_sample[self.input_col][i] = parsed_history.replace(
```
jonathangomesselman, line 175: Now I get it! I think it would be helpful to add a comment saying how this is overriding `user_col` - which makes us only have to call `user_col[i-1]` to get the full history.

jonathangomesselman (approved):
Looks great! 

One question to follow up is how fast this is? We for example are re-tokenizing the same strings quite a bit (i.e. the number of "turns" that a given turn appears in the history). But looks good to push up for now.

dataquality/integrations/seq2seq/formatters/chat.py
elboy3: So we're doing 1. right now. I personally think we can push 1. and have sarah try it today and then ask her what they do with their models and if they'd prefer 2, or even some other solution (i think our whole chat experience will have to mature in the near future)

are people okay with that?

dataquality/integrations/seq2seq/formatters/chat.py
```diff
@@ -2,8 +2,13 @@
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
+from transformers import AutoTokenizer, PreTrainedTokenizerFast
+
 from dataquality.integrations.seq2seq.formatters.base import BaseFormatter
 
+NEWLINE_CHAR = "\n\n\n"
+NEWLINE_TOKEN = "[NEWLINE]"
```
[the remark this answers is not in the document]
elboy3: added a comment

dataquality/integrations/seq2seq/formatters/chat.py
elboy3: rewrote it to use `user_inputs` instead of `formatted_sample[self.input_col]` which i think will be more clear, also added comments to the whole for loop. lmk what you think

elboy3:
@jonathangomesselman  totally agree, we should not have to tokenize, decode, and then tokenize again later. i'll start working on that today

jonathangomesselman:
>There's a few different ways this could be done:

Take the last 512, irrespective of turns.
Greedily consume the most recent turns until we get as close to 512.
I think for v1 maybe we should just do 2 and then take it to customers so that maybe they can customize it.

jonathangomesselman:
>There's a few different ways this could be done:
>Take the last 512, irrespective of turns.
Greedily consume the most recent turns until we get as close to 512.
I think for v1 maybe we should just do 2 and then take it to customers so that maybe they can customize it.

jonathangomesselman:
>There's a few different ways this could be done:
- Take the last 512, irrespective of turns.
- Greedily consume the most recent turns until we get as close to 512.
I think for v1 maybe we should just do 2 and then take it to customers so that maybe they can customize it.

@setu4993 So (2) is how I did it in the first mock run we presented. To me this feels a bit more natural than taking the last 512 irrespective of turns. I think for now we are going to push something and then get feedback. But overall I more naturally align with (2).

elboy3:
Hey @setu4993 @jonathangomesselman I actually just made a last little update to support Setu's flow #1 .. let me know what you think

dataquality/integrations/seq2seq/formatters/chat.py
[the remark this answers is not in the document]
setu4993: Just a note that before running this through a model, we'd need to update the model's embedding dimensions as well.

setu4993 (approved):
Looks good, let's 🚢!

PR #784, opened 2023-10-31
fix(chore): cleanup s2s offsets

elboy3:
user cleaner vaex pattern of adding to df and return full DF

we also materialize `input_cutoff` to speedup larger runs

### DIFF
--- a/dataquality/integrations/seq2seq/s2s_trainer.py
+++ b/dataquality/integrations/seq2seq/s2s_trainer.py
@@ -17,6 +17,7 @@ from transformers import (
 )
 
 import dataquality as dq
+from dataquality.exceptions import GalileoException
 from dataquality.integrations.seq2seq.hf import watch
 from dataquality.integrations.seq2seq.schema import (
     Seq2SeqGenerationConfig,
@@ -24,6 +25,24 @@ from dataquality.integrations.seq2seq.schema import (
 )
 from dataquality.schemas.split import Split
 from dataquality.utils.torch import cleanup_cuda
+
+
+def validate_cols(ds: Dataset, input_col: str, target_col: str) -> None:
+    """Validates that the input and target columns are in the dataset"""
+    msg = (
+        "{col} column {val} not found in dataset. "
+        "Please check the DatasetConfig to ensure the {col_name} is correct."
+        "If you are using a custom formatter, please ensure that the "
+        "{col_name} is being set correctly."
+    )
+    if input_col not in ds.column_names:
+        raise GalileoException(
+            msg.format(col="Input", val=input_col, col_name="input_col")
+        )
+    if target_col not in ds.column_names:
+        raise GalileoException(
+            msg.format(col="Target", val=target_col, col_name="target_col")
+        )
 
 
 def tokenize(
@@ -34,6 +53,8 @@ def tokenize(
     max_input_length: int,
     max_target_length: int,
 ) -> Dataset:
+    validate_cols(ds, input_col, target_col)
+
     def _tokenize(row: Dict) -> BatchEncoding:
         """Tokenize the input and outputs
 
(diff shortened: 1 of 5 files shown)


dataquality/utils/seq2seq/offsets.py
```diff
@@ -228,16 +231,23 @@ def _get_position_of_last_offset_input(texts: pa.array) -> np.ndarray:
...
 
-def get_cutoff_from_saved_offsets(df: DataFrame, col_name: str) -> np.ndarray:
+def add_target_cutoff_to_df(df: DataFrame, target_offsets_col: str) -> DataFrame:
     """
     Look at the last offset of the tokenized target to find the position of the last
     character of the target string that was used by the model.
     Note that typically the model does not use the entire target during teacher forcing
     and there is a cut-off point (for example 128 tokens, or 512 tokens, etc).
     """
+    df_copy = df.copy()
```
bogdan-galileo, line 250: what is the copy for ?

bogdan-galileo (approved):
LGTM

dataquality/utils/seq2seq/offsets.py
elboy3: I think it's just good practice in a helper fn with vaex dataframe when you're updating the DF. we probably would be fine without but i add it as a safety precaution, it can help avoid unexpected side effects and sometimes you might want to preserve the original df without df updates. can be helpful for testing too

dataquality/integrations/seq2seq/s2s_trainer.py
```diff
@@ -26,6 +27,24 @@
...
+        "{col} column {val} not found in dataset. "
+        "Please check the DatasetConfig to ensure the {col_name} is correct."
+        "If you are using a custom formatter, please ensure that the "
+        "{col_name} is being set correctly."
+    )
+    if input_col not in ds.column_names:
+        raise GalileoException(
+            msg.format(col="Input", val=input_col, col_name="input_col")
+        )
+    if target_col not in ds.column_names:
```
franz101, line 42: Maybe also check for target and then if both are missing throw that

dataquality/utils/seq2seq/offsets.py
franz101, line 1: how long is the for loop? Can it be parallelized or combined with a jit (Jax or numbs)

dataquality/integrations/seq2seq/s2s_trainer.py
elboy3: good call, updated!

dataquality/utils/seq2seq/offsets.py
```diff
@@ -191,32 +192,34 @@ def align_tokens_to_character_spans(
...
     """
     Find the cutoff point in string coresponding to the last token.
 
-    We tokenize the text and truncate after max_n_tokens tokens, i.e., we only keep the
-    first max_n_tokens tokens. To find the position in the text corresponding to the
+    We tokenize the text and truncate after max_tokens tokens, i.e., we only keep the
+    first max_tokens tokens. To find the position in the text corresponding to the
     last token we use the offset_mapping returned by the tokenizer.
     """
+    df_copy = df.copy()
```
franz101, line 214: The df can't be edited by reference? I assume is not too expensive on ram.

dataquality/utils/seq2seq/offsets.py
elboy3: i'd say that's outside of the scope of this PR but we can (and will) look into speed improvements for seq2seq when we do a robustification sprint

dataquality/utils/seq2seq/offsets.py
elboy3: it could be but i'm not sure it's best vaex practice, and returning new df is helpful for testing

franz101 (approved):
Looks good to me, how much slower is seq2seq with dq vs without dq

elboy3:
> Looks good to me, how much slower is seq2seq with dq vs without dq

As in what's the overhead of logging with galileo? great question, i think it depends on a few things, like if you do generation or not. we should do some testing without generation on large runs to know exactly the overhead

PR #787, opened 2023-11-01
fix: include metadata cols in get dataframe

elboy3:
https://app.shortcut.com/galileo/story/8671/dq-get-dataframe-doesn-t-support-metadata

See also changes in API https://github.com/rungalileo/api/pull/1488

### DIFF
--- a/dataquality/metrics.py
+++ b/dataquality/metrics.py
@@ -1,6 +1,7 @@
 import json
 import os
 import warnings
+from dataclasses import field
 from typing import Any, Dict, List, Optional, Tuple, Union
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


dataquality/metrics.py
```diff
@@ -305,6 +308,8 @@ def get_dataframe(
         (embeddings, probabilities etc), vaex will always be returned, because pandas
         cannot support multi-dimensional columns. Default True
     :param include_data_embs: Whether to include the off the shelf data embeddings
+    :param meta_cols: List of metadata columns to return in the dataframe. If "*"
+        is passed, return all metadata columns
```
bogdan-galileo, line 312: Is it that the list has to contain the string `*` ?

elboy3: Yeah, so ["*"] would include all metadata

bogdan-galileo:
```suggestion
        is included, return all metadata columns
```

bogdan-galileo (approved):
LGTM

setu4993 (approved):

PR #791, opened 2023-11-02
feat: add seq2seq task types

elboy3:
Add `seq2seq_completion` and `seq2seq_chat` to task types


dataquality/loggers/data_logger/__init__.py
```diff
@@ -19,5 +23,7 @@
     "text_ner",
     "object_detection",
     "BaseGalileoDataLogger",
-    "seq2seq",
+    "base",  # Import needed for seq2seq subclass discovery
```
setu4993, line 26: Couldn't we just import directly in those files? Feels a bit weird to have `base` being exported.

setu4993: If we do need it, maybe makes sense to call it `seq2seq_base`? Just at the import-level would be fine IMO:
```suggestion
from dataquality.loggers.data_logger.seq2seq import base as seq2seq_base
...
    "seq2seq_base",  # Import needed for seq2seq subclass discovery
```

setu4993 (approved):
Broadly looks good, left 1 note about export name.

dataquality/loggers/data_logger/__init__.py
elboy3: Sounds good, i'll rename to base!

PR #793, opened 2023-11-08
fix: remove unused labels

elboy3:
In doing a refactor I noticed that labels are never used in model loggers, so let's remove them!


setu4993 (approved):
🔥

PR #795, opened 2023-11-13
feat: decoder only rearc

jonathangomesselman:

***Overview***

This is a **BIG** PR that both adds Decoder-Only logging support and also dramatically changes the Seq2Seq logging framework. The main idea is to separate out the two different model architectures underneath the umbrella Seq2Seq task category. 

Now, each model architecture has its own `Formatter` class that handles specific formatting of e.g. `data`, `logits`, ... 

Although differences exist, we also leverage as much overlap as possible through a common `Seq2Seq` base class.

Because we are working very fast to push this work, testing is currently quite limited. A follow up PR will be dedicated to add complete testing and ensuring strong engineering design throughout!

This was branched off of the initial branch `feature/decoder-only-as-formatter`. Also see this branch `feature/decoder-only` for other experimental work that @jonathangomesselman is doing.


elboy3 (approved):
lets do some testing and then 🚀 🚀 🚀 🚀 🚀

PR #796, opened 2023-11-13
Feat/s2s add tokenizer support

bogdan-galileo:
Adding support for tokenizers of type Tokenizer (to support Cohere) https://app.shortcut.com/galileo/story/8963


bogdan-galileo:
Will clean up the history and start from main...

PR #797, opened 2023-11-14
chore: add support for Tokenizer

bogdan-galileo:
Adding support for tokenizers of type Tokenizer (to support Cohere) https://app.shortcut.com/galileo/story/8963

The class `from tokenizers import Tokenizer` is a performant implementation of lots of popular tokenizers https://github.com/huggingface/tokenizers. It is implemented in Rust with bindings in Python.

The class `from transformers import PreTrainedTokenizerFast` seems more popular as it integrates easier with transformers and many trained tokenizers have this type https://huggingface.co/docs/transformers/v4.33.2/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.

We first implemented support for PreTrainedTokenizerFast and require support for Tokenizer (in particular for Cohere). It turns out a Tokenizer can be converted into a PreTrainedTokenizerFast, so that's what we'll use for now since it allows minimal code change. We can think about a more general solution if we'll need to support more general tokenizers

### DIFF
--- a/dataquality/integrations/seq2seq/hf.py
+++ b/dataquality/integrations/seq2seq/hf.py
@@ -1,6 +1,7 @@
-from typing import List, Optional
+from typing import List, Optional, Union
 from warnings import warn
 
+from tokenizers import Tokenizer
 from transformers import GenerationConfig, PreTrainedModel, PreTrainedTokenizerFast
 
 from dataquality.exceptions import GalileoException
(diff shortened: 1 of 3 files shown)


bogdan-galileo:
The failing test is due to an ultralytics (for OD) upgrade that will be fixed here https://github.com/rungalileo/dataquality/pull/798

dataquality/integrations/seq2seq/hf.py
```diff
@@ -41,37 +47,41 @@ def set_tokenizer(
...
     task_type = get_task_type()
     assert task_type == TaskType.seq2seq, "This method is only supported for seq2seq"
+
+    tokenizer_dq = tokenizer
+    if isinstance(tokenizer, Tokenizer):
+        tokenizer_dq = PreTrainedTokenizerFast(tokenizer_object=tokenizer)
     assert isinstance(
-        tokenizer, PreTrainedTokenizerFast
+        tokenizer_dq, PreTrainedTokenizerFast
     ), "Tokenizer must be an instance of PreTrainedTokenizerFast"
```
elboy3, line 56: should we say "must be an instance of PreTrainedTokenizerFast or Tokenizer" now?

tests/integrations/seq2seq/test_hf.py
```diff
@@ -0,0 +1,43 @@
...
+from dataquality.integrations.seq2seq.hf import set_tokenizer
+from tests.conftest import TestSessionVariables, tokenizer_T5, tokenizer_T5_not_auto
+
+
+def test_set_tokenizer(
+    set_test_config: Callable,
+    cleanup_after_use: Callable,
+    test_session_vars: TestSessionVariables,
+) -> None:
+    """
```
elboy3, line 17: for best "pytest" practices I would honestly split this up into 3 really small tests rather than 1 test that tests 3 flows

elboy3 (approved):
love it! had 2 suggestions:
- one about assertion message
- one about more but smaller tests :)

PR #799, opened 2023-11-14
chore: Adding support for logging just logprobs from an API model

jonathangomesselman:

***PR Description***

This PR add support for users to log just `logprob` data as model outputs - i.e. not the traditional **logits** (logprobs are true log probabilities) and not over the full token vocabulary (we just get the logprob for the given token).

To remove the need for API / UI changes, we generate fake `top_k` logprob data of the following form - each token has `top_logprob` data:
```
[("---", -20), ("---", -20), ...]
```

Thankfully, we were able to reuse a lot of the logic we are already using for processing logits - essentially we just have `token_logprobs` and we mainly need to remove padding, etc.

E2E with API based logging notebook

***Demonstration***

See the example notebook `docs/notebooks/Seq2Seq_API_Logging_Flow.ipynb`

As a sample API integration we use the publicly available APIs from the `Cohere` completion model over the `billsum` dataset. Note that you will need a personal Cohere API key to use the notebook.


jonathangomesselman:
https://app.shortcut.com/galileo/story/9008/dq-add-support-for-logging-just-logprob-data

dataquality/integrations/seq2seq/hf.py
```diff
@@ -123,6 +124,13 @@ def watch(
     ), "model must be an instance of transformers PreTrainedModel"
     assert model.can_generate(), "model must contain a `generate` method for seq2seq"
 
+    assert model_type in [
+        Seq2SeqModelTypes.encoder_decoder,
+        Seq2SeqModelTypes.decoder_only,
+    ], "model_type must be one of `encoder_decoder` or `decoder_only`"
+    if model_type == Seq2SeqModelTypes.decoder_only:
+        seq2seq_logger_config.model_type = Seq2SeqModelTypes.decoder_only
```
bogdan-galileo, line 132: It seems to already be `decoder_only` by default ? Maybe we should have no if at all and just something like `seq2seq_logger_config.model_type = model_type` (or maybe it's `Seq2SeqModelTypes[model_type]`)

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -75,27 +76,36 @@ def validate_and_format(self) -> None:
         """Validate the lengths, calculate token level dep, extract GT probs"""
         self.embs = self._convert_tensor_ndarray(self.embs)
         self.logits = self._convert_tensor_ndarray(self.logits)
-        self.logprobs = self._convert_tensor_ndarray(self.logprobs)
+        # TODO THIS SHOULD BE LOGPROBS
```
bogdan-galileo, line 79: Something to do here ?

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -75,27 +76,36 @@ def validate_and_format(self) -> None:
...
-        ) = self.process_logprobs(
-            self.ids, self.logits  # type: ignore
-        )
 
-    def process_logprobs(
+        if self.probs is not None:
+            (self.token_logprobs, self.top_logprobs) = self.process_logprobs(
+                self.ids, self.probs
+            )
+        else:
```
bogdan-galileo, line 97: Can self.probs ever be None though ? Wouldn't `self.probs = self._convert_tensor_ndarray(self.probs)` break if it's the case ?

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -150,6 +162,63 @@ def process_logprobs(
...
+        for sample_id, sample_logprobs in zip(batch_ids, batch_logprobs):
+            # Note that with API based models the logprob data is
+            # already shifted / aligned.
+            response_labels, sample_response_logprobs = self.formatter.format_sample(
+                sample_id, sample_logprobs, self.split_key, shift_labels=False
+            )
+
+            # Add fake top loprobs
+            sample_top_logprobs: List[List[Tuple[str, float]]] = [
+                [("---", -20)] * TOP_K
```
bogdan-galileo, line 211: Changing the -20 to -1e10 ?

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
bogdan-galileo: Maybe we can move the _convert inside the if ? Both for probs and logits ?

bogdan-galileo: And that assert len(self.ids) too ?

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -150,6 +162,63 @@ def process_logprobs(
...
+        batch_top_logprobs = []
+        for sample_id, sample_logprobs in zip(batch_ids, batch_logprobs):
+            # Note that with API based models the logprob data is
+            # already shifted / aligned.
+            response_labels, sample_response_logprobs = self.formatter.format_sample(
+                sample_id, sample_logprobs, self.split_key, shift_labels=False
+            )
+
+            # Add fake top loprobs
+            sample_top_logprobs: List[List[Tuple[str, float]]] = [
```
bogdan-galileo, line 210: It would seem nicer to me if `sample_top_logprobs` would be None in this case, or an empty array, or something obvious to check and the UI takes care of populating fake stuff. Maybe that's too much work for now and would require too many changes downstream to accomodate the new column format

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -174,3 +243,20 @@ def _get_data_dict(self) -> Dict:
...
+        work with logprobs. One reason for this is that the logsoftmax
+        function takes advantage of the logsumexp "trick" to compute a
+        numerically stable version of log(softmax(x)).
+        """
+        # axis ensures that in a matrix of probs with dims num_samples x num_classes
+        # we take the softmax for each sample
+        if not isinstance(sample_logits, np.ndarray):
+            sample_logits = self._convert_tensor_ndarray(sample_logits)
+
+        return log_softmax(sample_logits, axis=-1)
```
bogdan-galileo, line 262: Is this slow ? Should it be done on GPU with Torch ?

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
```diff
@@ -150,6 +162,63 @@ def process_logprobs(
...
+                sample_id, sample_logprobs, self.split_key, shift_labels=False
+            )
+
+            # Add fake top loprobs
+            sample_top_logprobs: List[List[Tuple[str, float]]] = [
+                [("---", -20)] * TOP_K
+            ] * len(response_labels)
+
+            batch_token_logprobs.append(sample_response_logprobs)
+            batch_top_logprobs.append(sample_top_logprobs)
```
bogdan-galileo, line 215: Alternatively we can construct `batch_top_logprobs` after the for loop in 1 shot, using the dim of `batch_token_logprobs` (not 100% sure ?)

dataquality/loggers/logger_config/seq2seq/seq2seq_base.py
```diff
@@ -18,7 +18,9 @@ class Seq2SeqLoggerConfig(BaseLoggerConfig):
     model: Optional[PreTrainedModel] = None
     generation_config: Optional[GenerationConfig] = None
     generation_splits: Set[Split] = set()
-    model_type: Seq2SeqModelTypes = Seq2SeqModelTypes.encoder_decoder
+    model_type: Seq2SeqModelTypes = (
+        Seq2SeqModelTypes.decoder_only
+    )  # Seq2SeqModelTypes.encoder_decoder
```
elboy3, line 23: did you mean to revert to default encoder only?

dataquality/integrations/seq2seq/hf.py
```diff
@@ -101,6 +101,7 @@ def watch(
     model: PreTrainedModel,
     tokenizer: PreTrainedTokenizerFast,
     generation_config: GenerationConfig,
+    model_type: Optional[str] = Seq2SeqModelTypes.encoder_decoder.value,
```
elboy3, line 104: how do you feel about making this not optional.... enforcing that they always are explicit about which model type they want?

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: I think this is going to be addressed now in Elliott's PR

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
jonathangomesselman: Oh interesting! I don't think it is None but I think it will be an empty list.

jonathangomesselman: Changing this check to `if len(self.probs) != 0`

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
jonathangomesselman: Yeah -1e10 leads to a ridiculously small value that was breaking the UI lol. Even -20 is e^-20.

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
jonathangomesselman: May default to @elboy3 - Originally I wanted to have that kind of logic, but I wasn't sure if it added unnecessary complication on the backend

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
jonathangomesselman: Definitely lol - but maybe for a follow up PR.

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
jonathangomesselman: True but we add to `batch_token_logropbs` in the loop so may as well keep it here?

dataquality/integrations/seq2seq/hf.py
jonathangomesselman: I like this!

dataquality/loggers/model_logger/seq2seq/seq2seq_base.py
bogdan-galileo: -20 is good, as you said, the prob is 1/e^20

elboy3 (approved):
🚀 🚀 🚀  looks good! updated a few small things to get linting passing

PR #806, opened 2023-11-16
Improved progress logging when processing data

jonathangomesselman:

Adds progress logging for two of the longer steps during `data_logging`. Specifically, tokenization and alignment is the bottleneck during data processing. So now we add the following progress bar logging steps:
1. `Identifying `response_template` to isolate response tokens`: When we loop through samples and locate the `response_template`
2. `Aligning string characters with tokenizer representation`: This is generally the biggest bottleneck, since we currently decode the `response_tokens` then re-tokenize them to get the offsets 😬. 

Overall these progress bars just make it feel a lot less daunting!

I also added a small change to the   `extract_tokenized_responses` that breaks early when we have found the response template in reverse order. This is actually important in case they have multiple instances of the response template. 

***Demonstration***

https://github.com/rungalileo/dataquality/assets/29869344/aa76ab56-8604-45df-aed0-0b5761d60773

### DIFF
--- a/dataquality/utils/seq2seq/decoder_only.py
+++ b/dataquality/utils/seq2seq/decoder_only.py
@@ -1,12 +1,12 @@
-from typing import List, Optional
+from typing import List
 from warnings import warn
-from tqdm.auto import tqdm
 
 import numpy as np
+from tqdm.auto import tqdm
 
 
 def extract_tokenized_responses(
-    tokenized_formatted_prompts: List[List[int]], response_template: Optional[List[int]]
+    tokenized_formatted_prompts: List[List[int]], response_template: List[int]
 ) -> List[List[int]]:
     """Extracts the tokenized responses from the formatted prompts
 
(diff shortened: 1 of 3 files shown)


bogdan-galileo (approved):
Awesome!

dataquality/utils/seq2seq/decoder_only.py
```diff
@@ -42,6 +43,7 @@ def extract_tokenized_responses(
                 == response_template
             ):
                 response_token_ids_start_idx = match_idx
+                break
```
elboy3, line 46: can you leave a comment why we are breaking? (it's cause we can stop after we found a match, right?)

elboy3 (approved):
if you bump the patch version in `__init__` as well we can get this out to cohere folks

dataquality/utils/seq2seq/decoder_only.py
jonathangomesselman: Yeah since we are reverse searching for matches of the just the first token, if we find the whole `response_template` we can stop - and actually should!

PR #807, opened 2023-11-20
feat: configure data emb column

elboy3:
Allows the user to configure which column they want to calculate data embeddings on https://app.shortcut.com/galileo/story/9156/dq-allow-data-embs-column-to-be-configurable

Related to https://github.com/rungalileo/dataquality/pull/808


setu4993 (approved):
This is neat. I like that our invocations for various fns are now very explicit in the args.

PR #808, opened 2023-11-20
chore(refactor): rename s2s text to input label to target

elboy3:
For S2S instead of mapping to `text` and `label` for S2S processing and renaming to `input` and `target` at the end, let's just map to `input` and `target` as the column names. This will be helpful with configuring which column to apply certain algorithms on, such as data embs

## Testing
All existing tests still pass
✅  Have created a few e2e internal tests with alpaca and chat data to confirm the changes as well 

## What else
This simplifies the data embs configurability change https://github.com/rungalileo/dataquality/pull/807/files

### DIFF
--- a/dataquality/schemas/seq2seq.py
+++ b/dataquality/schemas/seq2seq.py
@@ -4,7 +4,6 @@ from typing import List, Set, Tuple
 
 import numpy as np
 import pyarrow as pa
-from vaex import DataFrame
 
 # Defines the format schema for storing top_logprobs as a
 # pyarrow List of List of Tuples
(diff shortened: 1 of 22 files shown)


dataquality/schemas/seq2seq.py
```diff
@@ -39,27 +36,6 @@ class Seq2SeqInputCols(str, Enum):
...
-
-        return df
-
-    @classmethod
-    def set_target(cls, df: DataFrame) -> DataFrame:
-        """Sets the target output column for the dataframe"""
-        if cls.label.value in df.get_column_names():
-            df.rename(cls.label.value, cls.target.value)
-
-        return df
```
elboy3, line 61: so nice to have these gone 👋

setu4993: 🔥

setu4993 (approved):
Love this!

PR #809, opened 2023-11-22
feat: s2s add high dep tokens

bogdan-galileo:
Goal is to compute high dep tokens https://app.shortcut.com/galileo/story/9232/dq-save-nessec-column-for-high-dep-tokens?vc_group_by=day&ct_workflow=all&cf_workflow=500000005

The DQ work requires saving the target tokens. We later groupby them (and the ID is not enough as we need the tokens for the UI).

Decoding the target token ids is fast enough in my opinion. Here it's taking ~10 sec for 10K rows and ~1M tokens 
Screenshot 2023-11-28 at 10 17 41 AM.

### DIFF
--- a/dataquality/loggers/data_logger/seq2seq/formatters.py
+++ b/dataquality/loggers/data_logger/seq2seq/formatters.py
@@ -102,12 +102,11 @@ class EncoderDecoderDataFormatter(BaseSeq2SeqDataFormatter):
         """Further validation for Encoder-Decoder
 
         For Encoder-Decoder we need to:
-            - Save the offsets and positions of the tokenized labels
-                Allows us to extract token level information from the full
-                sample text
-            - Save the target token id
-                Equivalent to the "ground truth", allows us to get logprob per
-                token later on
+            - Save the target token ids: Equivalent to ground truth, it allows us to
+                compare with the predictions and get perplexity and DEP scores
+            - Save the target tokens: Decoding of the ids, to identify the tokens
+            - Save the offsets and positions of the target tokens: allows us to extract
+                token level information and align the tokens with the full sample text
 
         We achieve this by:
             - Tokenize the target texts using `max_target_tokens`
@@ -124,13 +123,15 @@ class EncoderDecoderDataFormatter(BaseSeq2SeqDataFormatter):
             max_length=max_target_tokens,
             truncation=True,
         )
-        tokenized_labels = encoded_data["input_ids"]
+        token_label_ids = encoded_data["input_ids"]
+        data_logger.token_label_str = tokenizer.batch_decode(token_label_ids)
+
         aligned_data = align_tokens_to_character_spans(encoded_data["offset_mapping"])
         data_logger.token_label_offsets = aligned_data.token_label_offsets
         data_logger.token_label_positions = aligned_data.token_label_positions
 
-        # Save the tokenized response labels for each samples
-        id_to_tokens = dict(zip(data_logger.ids, tokenized_labels))
+        # Save the token_ids in the config (to share with the model logger)
+        id_to_tokens = dict(zip(data_logger.ids, token_label_ids))
         self.logger_config.id_to_tokens[data_logger.split_key].update(id_to_tokens)
 
     def set_input_cutoff(self, df: DataFrame) -> DataFrame:
--- a/dataquality/loggers/data_logger/seq2seq/formatters.py
+++ b/dataquality/loggers/data_logger/seq2seq/formatters.py
@@ -266,15 +266,13 @@ class DecoderOnlyDataFormatter(BaseSeq2SeqDataFormatter):
             desc="Aligning string characters with tokenizer representation",
         ):
             # Detokenize to save the token_str in the df (for ex for high DEP tokens)
-            # Need to decode row by row otherwise each row is joined into one string
-            data_logger.token_label_str = [
+            data_logger.token_label_str.append(
                 tokenizer.batch_decode(
-                    row,
+                    token_label_ids,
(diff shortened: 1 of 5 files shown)


dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -124,13 +123,15 @@ def format_text(
             max_length=max_target_tokens,
             truncation=True,
         )
-        tokenized_labels = encoded_data["input_ids"]
+        token_label_ids = encoded_data["input_ids"]
+        data_logger.token_label_str = tokenizer.batch_decode(token_label_ids)
```
elboy3, line 127: so batch decode is on scale of a few seconds? that's nice

elboy3:
Have we created any runs in dev minio as a little test?

dataquality/loggers/data_logger/seq2seq/formatters.py
bogdan-galileo: Yeah. If it ever start to take longer at some point (larger datasets, decode not as good, etc) we could also extract the token string values using the text + the offsets, it's immediate (~50 times faster).

I just don't like it as much as we'd have to clean up the string from " " and "\n" and probably other things
Screenshot 2023-11-28 at 10 24 09 AM

bogdan-galileo:
> Have we created any runs in dev minio as a little test?

Yeah just tested, looks good. After processing the length of the tokens aligns with the length of tokens deps too 
Screenshot 2023-11-28 at 10 45 18 AM

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -118,19 +117,31 @@ def format_text(
         """
         targets = text
         max_target_tokens = max_tokens
+        use_special_tokens = True  # use this var to align encoding and decoding
```
jonathangomesselman, line 120: Why do we create a var here and not just call pass in `True`

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -118,19 +117,31 @@ def format_text(
...
+            add_special_tokens=use_special_tokens,
         )
-        tokenized_labels = encoded_data["input_ids"]
+        token_label_ids = encoded_data["input_ids"]
+        # Need to decode row by row otherwise each row is joined into one string
+        data_logger.token_label_str = [
+            tokenizer.batch_decode(
+                row,
+                skip_special_tokens=not use_special_tokens,
+                clean_up_tokenization_spaces=True,
```
jonathangomesselman, line 134: What does `clean_up_tokenization_spaces` do again? I have seen it but never fully understood it.

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -247,14 +260,25 @@ def format_text(
...
         # Decode then re-tokenize just the response labels to get correct offsets
-        for tokenized_response in tqdm(
+        for token_label_ids in tqdm(
             tokenized_labels,
             leave=False,
             desc="Aligning string characters with tokenizer representation",
         ):
+            # Detokenize to save the token_str in the df (for ex for high DEP tokens)
+            # Need to decode row by row otherwise each row is joined into one string
+            data_logger.token_label_str = [
```
jonathangomesselman, line 270: I don't think this is going to work. The result of this will be just saving the de-tokenized tokens for the last sample. For decoder only we loop over each sample individually, since we need to isolate the response template. You probably want to do this outside the loop!

jonathangomesselman (changes requested):
Looks great! I left some feedback in the `Decoder-Only` section. I think that you may need to move the location of where you do the decoding to properly decode all of the samples.

dataquality/loggers/data_logger/seq2seq/formatters.py
bogdan-galileo: I also use it below in the decoding, so it's to make sure that they align so that the token_ids and token_str have the same lengths

dataquality/loggers/data_logger/seq2seq/formatters.py
bogdan-galileo: Supposedly to remove spaces added during tokenization https://discuss.huggingface.co/t/what-does-the-parameter-clean-up-tokenization-spaces-do-in-the-tokenizer-decode-function/17399. But I've never seen those spaces with the T5 tokenizer, and in the decoder only notebook, the tokenizer adds those spaces at the beginninng of each word, and using True does not remove the spaces!

Still I think it cleans up a bit the token string in some cases.

dataquality/loggers/data_logger/seq2seq/formatters.py
bogdan-galileo: Fixed and did a run with decoder only, the data in minio aligns

Screenshot 2023-11-29 at 11 40 13 AM

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -247,14 +260,23 @@ def format_text(
...
 
         # Decode then re-tokenize just the response labels to get correct offsets
-        for tokenized_response in tqdm(
+        for token_label_ids in tqdm(
             tokenized_labels,
             leave=False,
             desc="Aligning string characters with tokenizer representation",
         ):
+            # Detokenize to save the token_str in the df (for ex for high DEP tokens)
+            data_logger.token_label_str.append(
```
jonathangomesselman, line 269: Nice!

jonathangomesselman (approved):
Looks good to me!

PR #810, opened 2023-11-28
Clean up data `formaters`, removing need for passing `Seq2SeqDataLogger`

jonathangomesselman:
***PR Description***

One of the current design choices for the `data_logger` formatters, was to pass the actual `Seq2SeqDataLogger` to the `.format()` function. This was nice for dev speed, but is not the nicest for three primary reason (that I see):
1. It can cause annoying trickiness around avoiding circular imports. For example, I am working on the generation PR for Decoder-Only models, and I want to use `data_formatters`; however, when I try to do so, I get circular import errors.
2. Overall it seems like not the greatest python "style".
3. It is not as necessary as we maybe initially thought.

To elaborate on point (3), you will see in this PR that we basically need to address two main issues when removing the dependance on having the actual `Seq2SeqDataLogger`:
1. Saving information to the `LoggerConfig`: Accomplished by passing in two additional data fields to the `format` function: a) `ids` and b) `split_key`
2. Assigning offsets and positions: Accomplished by returning the "formatted" `AlignedData` object. 

With these two fairly simple extensions we are able to simplify the `format(.)` function!

Note: One addition that we have to make is extending the `AlignedTokenData` class to have an `append(.)` function. This is necessary because for DecoderOnly models we process the aligned token data one sample at a time. The function we add is simple and follows this structure:
```
# The AlignedTokenData object that we are building up sample by sample
aligned_data: AlignedTokenData

# Aligned token data for a *single* sample - i.e. batch dim == 1
sample_aligned_token_data: AlignedTokenData

# This will append the `offset` and `position` data from the sample
# to the internal fields of `aligned_data`.
aligned_data.append(sample_aligned_token_data)
```

### DIFF
--- a/dataquality/loggers/data_logger/seq2seq/formatters.py
+++ b/dataquality/loggers/data_logger/seq2seq/formatters.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Dict, List, Optional, Type, Tuple
+from typing import Dict, List, Optional, Tuple, Type
 
 from tqdm.auto import tqdm
 from transformers import PreTrainedTokenizerFast
@@ -137,13 +137,15 @@ class EncoderDecoderDataFormatter(BaseSeq2SeqDataFormatter):
             for row in token_label_ids
         ]
 
-        aligned_data = align_tokens_to_character_spans(encoded_data["offset_mapping"])
+        batch_aligned_data = align_tokens_to_character_spans(
+            encoded_data["offset_mapping"]
+        )
 
         # Save the token_ids in the config (to share with the model logger)
         id_to_tokens = dict(zip(ids, token_label_ids))
         self.logger_config.id_to_tokens[split_key].update(id_to_tokens)
 
-        return aligned_data, token_label_str
+        return batch_aligned_data, token_label_str
 
     def set_input_cutoff(self, df: DataFrame) -> DataFrame:
         """Calculate the cutoff index for the input strings.
@@ -262,7 +264,7 @@ class DecoderOnlyDataFormatter(BaseSeq2SeqDataFormatter):
         )
 
         # Empty initialization
-        aligned_data = AlignedTokenData([], [])
+        batch_aligned_data = AlignedTokenData([], [])
         token_label_str = []
         # Decode then re-tokenize just the response labels to get correct offsets
         for token_label_ids in tqdm(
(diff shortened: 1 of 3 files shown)


jonathangomesselman:
https://app.shortcut.com/galileo/story/9349

bogdan-galileo (approved):
LGTM

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -246,37 +248,36 @@ def format_text(
...
 
         # Save the length of the formatted prompt - used later to remove padding
         formatted_prompt_lengths = [
             len(prompt) for prompt in tokenized_formatted_prompts
         ]
-        id_to_formatted_prompt_length = dict(
-            zip(data_logger.ids, formatted_prompt_lengths)
-        )
-        self.logger_config.id_to_formatted_prompt_length[data_logger.split_key].update(
+        id_to_formatted_prompt_length = dict(zip(ids, formatted_prompt_lengths))
```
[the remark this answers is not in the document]
jonathangomesselman: I don't hate it. May leave it since this is for some reason the style that has been used in DQ. Don't want to rock the boat.

dataquality/loggers/data_logger/seq2seq/formatters.py
```diff
@@ -246,37 +248,36 @@ def format_text(
...
+        # Empty initialization
+        aligned_data = AlignedTokenData([], [])
         # Decode then re-tokenize just the response labels to get correct offsets
         for tokenized_response in tqdm(
             tokenized_labels,
             leave=False,
             desc="Aligning string characters with tokenizer representation",
         ):
-            aligned_data = align_response_tokens_to_character_spans(
+            sample_aligned_data = align_response_tokens_to_character_spans(
```
dcaustin33, line 259:
```suggestion
            response_aligned_data = align_response_tokens_to_character_spans(
```

dataquality/schemas/seq2seq.py
```diff
@@ -71,6 +71,22 @@ class AlignedTokenData:
...
+    def append(self, sample_aligned_token_data: "AlignedTokenData") -> None:
+        """Append offsets and positions for a *single* sample
+
+        Assumes that `sample_aligned_token_data` holds alignment info for
+        a single data sample. As such, when appending to `token_label_offsets`
+        and `token_label_positions` we remove the "batch" dimensions respectively.
+            e.g.
+            >> sample_aligned_token_data.token_label_offsets[0]
+        """
+        self.token_label_offsets.append(
```
dcaustin33, line 83: Would add an assert here to make sure the length of `sample_aligned_token_data.token_label_offsets[0]` is always 1 so people do not use it with multiple samples..

dcaustin33 (approved):
Would be good to add some failure cases whenever you do tests!

dataquality/schemas/seq2seq.py
jonathangomesselman: Good call!

PR #812, opened 2023-11-29
feat: add warnings in s2s for bad inputs

bogdan-galileo:
Refactoring/Robustification https://app.shortcut.com/galileo/story/9394/add-warning-for-bad-inputs-response-template-and-data-emb-col

- If`response_template` is not passed as a list of int (token ids), raise an explicit exception
- If `data_emb_col` cannot be found in the df, first fallback to default cols text and target, and raise an exception if still not found (should not happen though)

Current error if data_embs_col is not specified (or specified wrong)

Screenshot 2023-11-30 at 11 42 36 AM

### DIFF
--- a/dataquality/integrations/seq2seq/core.py
+++ b/dataquality/integrations/seq2seq/core.py
@@ -121,8 +121,8 @@ def watch(
     """Seq2seq only. Log model generations for your run
 
     Iterates over a given dataset and logs the generations for each sample.
-    `model` must be an instance of transformers PreTrainedModel and have a `generate`
-    method.
+    To generate outputs, a model of instance of transformers PreTrainedModel must be
+    given and it must have a `generate` method.
 
     Unlike other watch functions, in this one we are just registering the model
     and generation config and not attaching any hooks to the model. We call it 'watch'
@@ -135,25 +135,24 @@ def watch(
         max_target_tokens=max_target_tokens,
     )
 
-    if model:
-        assert isinstance(
-            model, PreTrainedModel
-        ), "model must be an instance of transformers PreTrainedModel"
-        assert (
-            model.can_generate()
-        ), "model must contain a `generate` method for seq2seq"
-
-    if model_type == Seq2SeqModelType.decoder_only and not response_template:
-        raise GalileoException(
-            "You must specify a `response_template` when using Decoder-Only models."
-            " This is necessary to internally isolate the target response tokens."
-        )
-
-    if model_type == Seq2SeqModelType.encoder_decoder and response_template:
+    if model_type == Seq2SeqModelType.decoder_only:
+        if response_template is None:
+            raise GalileoException(
+                "You must specify a `response_template` when using Decoder-Only models."
+                " This is necessary to internally isolate the target response tokens."
+            )
+        elif not isinstance(response_template, list) or not all(
+            isinstance(token, int) for token in response_template
+        ):
+            raise GalileoException(
+                "The response template must already be tokenized and be a list of ints."
+            )
+    elif model_type == Seq2SeqModelType.encoder_decoder and response_template:
         warn(
             "The argument response_template is only used when working with "
             "DecoderOnly models. This value will be ignored."
         )
+
     seq2seq_logger_config.response_template = response_template
     seq2seq_logger_config.model = model
     seq2seq_logger_config.generation_config = generation_config
@@ -169,5 +168,14 @@ def watch(
             continue
 
         generation_splits_set.add(Split[split])
+
+    # A model of the correct type is required if we need to generate
+    if generation_splits:
+        assert isinstance(
+            model, PreTrainedModel
+        ), "model must be an instance of transformers PreTrainedModel"
+        assert (
+            model.can_generate()
+        ), "model must contain a `generate` method for seq2seq"
 
     seq2seq_logger_config.generation_splits = generation_splits_set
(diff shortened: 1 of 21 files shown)


dataquality/integrations/seq2seq/core.py
```diff
@@ -135,25 +135,24 @@ def watch(
         max_target_tokens=max_target_tokens,
     )
 
-    if model:
```
bogdan-galileo, line 138: model is only necessary if we have geenration. Adding the extra check that if we have generation_splits, we need a model

tests/integrations/seq2seq/test_core.py
```diff
@@ -56,10 +57,9 @@ def test_set_tokenizer_other(
     )
     with pytest.raises(ValueError) as e:
         set_tokenizer(tokenizer_T5_not_auto, "encoder_decoder")
-        assert str(e.value) == (
-            "The tokenizer must be an instance of PreTrainedTokenizerFast "
-            "or Tokenizer"
-        )
+    assert str(e.value) == (
```
bogdan-galileo, line 60: This assert is not tested in the `with` clause, it has to be done outside (we can put an `assert False` in the `with` clause and the test passes!)

jonathangomesselman: Interesting!

jonathangomesselman: So these tests were not passing before?

tests/loggers/test_seq2seq.py
```diff
@@ -558,3 +558,37 @@ def test_create_and_upload_data_embs(
...
+            "input": [input_1, input_2],
+            "target": [target_1, target_2],
+        }
+    )
+    data_logger = Seq2SeqDataLogger()
+    data_logger.log_dataset(ds, text="input", label="target", split="training")
+
+    df = vaex.open(f"{data_logger.input_data_path}/**/data*.arrow")
+
+    with pytest.raises(GalileoException) as e:
```
jonathangomesselman, line 589: I don't know if I quite get this test. Like what is the column "not_input"? How does the user do something with this?

jonathangomesselman (commented):
Love the start of adding tests! I had one question about the data embs question

tests/loggers/test_seq2seq.py
bogdan-galileo: Yeah good point. I think the typical scenario is when the user does not specify it (instead of specifying it wrong) and then we use the default one set to `text` which is not found. Updating the test to the case where we don't specify it

tests/integrations/seq2seq/test_core.py
bogdan-galileo: They were passing, it's just that this line was not taken into account. We can literally replace it with `assert False` and the test would still pass. To make it count we have to put it outside of the `with` clause

bogdan-galileo: Screenshot 2023-11-30 at 11 20 14 AM

tests/loggers/test_seq2seq.py
jonathangomesselman: Ah so the user can set what column they want to compute data embeds on?

tests/integrations/seq2seq/test_core.py
jonathangomesselman: Cool did not know this!

tests/loggers/test_seq2seq.py
bogdan-galileo: O, I might not have answered what you asked.

The user calls `dq.finish(data_embs_col=data_embs_col)` which then ends up calling `create_data_embs_df(df, text_col=date_embs_col)`. There are multiple possible paths to get to `create_data_embs_df()` though (depending if you have cuML or not), so I thought I'd just test `create_data_embs_df`.

We can also test by calling `dq.finish` (or maybe the intermediate `upload()`), but then we also have to log model outputs so it made the test more complicated. Let me know what you think

jonathangomesselman: I didn't realize that! I think I understand a bit better now.

bogdan-galileo: Yeah exactly. It's a change that @elboy3 made not long ago. You have to specify `dq.finish(date_embs_col="target")` or whatever column you want. If not specified, we use `text` by default, and then it throws a weird cryptic error. 
 
Which is why I added this clearer error message so they understand. I'll add one sentence to indicate to rerun dq.finish

jonathangomesselman: I somehow thought though that we rename the some column always to `text` probably the `input`. So shouldn't it always exist.

jonathangomesselman: I think because in Seq2Seq we rename to `Input` and `Target`!

bogdan-galileo: Good point. Using the user column first, then trying `text` then `input`

dataquality/integrations/seq2seq/core.py
```diff
@@ -121,8 +121,8 @@ def watch(
     """Seq2seq only. Log model generations for your run
 
     Iterates over a given dataset and logs the generations for each sample.
-    `model` must be an instance of transformers PreTrainedModel and have a `generate`
-    method.
+    To generate outputs, a model of instance of transformers PreTrainedModel must be
+    given and it must have a `generate` method.
```
elboy3, line 125:
```suggestion
    To generate outputs, a model that is an instance of transformers PreTrainedModel
     must be given and it must have a `generate` method.
```

elboy3 (approved):
looks good! made a few small tweaks

PR #814, opened 2023-12-04
fix: disallow emb cols as metadata

elboy3:
https://app.shortcut.com/galileo/story/9579/dq-adding-x-y-data-x-or-data-y-as-metadata-succeeds-but-should-not

Customer ran into issues when logging x as metadata


setu4993 (approved):
Wow, fun debugging I'm sure!

elboy3:
@setu4993 yes we were getting quite an odd error on the runners side!

PR #818, opened 2023-12-13
feat: token refresh

elboy3:
## Shortcut Ticket
https://app.shortcut.com/galileo/story/10174/dq-api-client-token-refresh

## Drew Inspiration from
We should have an automatic token refresh similar to LLM Monitor and Prompt: https://github.com/rungalileo/llm-monitor/blob/1a42e5275dc5d4bb6b11219284269f44b02cab3b/llm_monitor/utils/api_client.py#L72-L89

## Testing
https://fmpm.dev/mocking-auth0-tokens

## Local example

My local token was expired but it used the local env to refresh succesfully.
```
>>> import os
>>> import jwt
>>> from time import time
>>> from dataquality.core._config import config

>>> os.environ["GALILEO_USERNAME"] = "<redacted>"
>>> os.environ["GALILEO_PASSWORD"] = "<redacted>"
>>> token = config.token
>>> exp = jwt.decode(token, options={"verify_signature": False})
>>> exp
{'sub': '<redacted>', 'id': '<redacted>', 'exp': 1700681062}
>>> time() > exp["exp"]
True

>>> ApiClient().get_current_user()
<response redacted>  # response succesful with refresh token
```

### DIFF
--- a/dataquality/core/auth.py
+++ b/dataquality/core/auth.py
@@ -15,32 +15,12 @@ api_client = ApiClient()
 
 
 class _Auth:
-    def email_login(self) -> None:
-        username = os.getenv("GALILEO_USERNAME")
-        password = os.getenv("GALILEO_PASSWORD")
-        res = requests.post(
-            f"{config.api_url}/login",
-            data={
-                "username": username,
-                "password": password,
-                "auth_method": "email",
-            },
-            headers={"X-Galileo-Request-Source": "dataquality_python_client"},
-        )
-        if res.status_code != 200:
-            raise GalileoException(
-                (
-                    f"Issue authenticating: {res.json()['detail']} "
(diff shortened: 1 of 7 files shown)


dataquality/core/auth.py
```diff
@@ -15,32 +12,12 @@
...
-                (
-                    f"Issue authenticating: {res.json()['detail']} "
-                    "If you need to reset your password, "
-                    f"go to {config.api_url.replace('api', 'console')}/forgot-password"
-                )
-            )
-
-        access_token = res.json().get("access_token")
-        config.token = access_token
-        config.update_file_config()
```
elboy3, line 41: i really don't like that we did this in auth

dmallon: ➕

PR #820, opened 2023-12-15
chore(deps): Update to `pydantic` v2

franz101:
Shortcut: https://app.shortcut.com/galileo/story/8757/update-dataquality-to-pydantic-v2

The last Pydantic v1 package is also going to v2 🚀.

### DIFF
--- a/tests/core/test_report.py
+++ b/tests/core/test_report.py
@@ -4,6 +4,7 @@ from uuid import uuid4
 
 import freezegun
 import vaex
+from pytest import raises
 
 from dataquality import (
     AggregateFunction,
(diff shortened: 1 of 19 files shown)


tests/core/test_report.py
```diff
@@ -202,8 +203,8 @@ def test_build_run_report_e2e(
             },
         ],
     }
-    mock_notify_email.assert_called_once_with(
-        expected_report_data, "run_report", ["foo@bar.com"]
-    )
+
+    # with pytest.raises(ValueError) as e:
+    # mock_notify_email(expected_report_data, "run_report", ["foo@bar.com"])
```
setu4993, line 208: Curious why this is commented out?

setu4993 (commented):
Mostly looks good. What's the issue with the deps?

tests/core/test_report.py
franz101: Ah i think i was debugging something. Good catch

franz101:
> Mostly looks good. What's the issue with the deps?

Hmm, just tried to run some tests and couldn't find it I think it was spacy or some dependency. Most tests are working too.  Will move the PR to draft right now as I'm focusing on prompt and monitoring

setu4993 (approved):

elboy3 (approved):
Let's also bump minor and then merge! 🚀

PR #821, opened 2023-12-15
Fix misalignment between token offsets returned from the API and samples in the UI 

jonathangomesselman:
***Issue Description***

Right now, for Decoder-Only models, there is a small mis-alignment bug between what is processed by the DQ client and what we see in the UI. Essentially, there is a small disconnect (for many situations) between what the model "sees" (as input and returns alignment / token data on) and <Target> string that we display in the UI.

**Current Flow**
- User logs <Target> string column - this is what will be displayed in the UI and what we expect to compute aligned token info for. However, we cannot directly compute alignment / token info since decoder only models process the full <Input> + <Target> through what we call the `formatted_prompt`
- User logs `formatted_prompt`
  - Locate response template
  - Slice off all characters / tokens following the response template
  - These tokens (and their direct string representation) is what the model is seeing and what we use to compute **token alignment** and **token likelihoods**.

From this flow, we need the logged <Target> text and the *sliced response text* (from formatted_prompt) to EXACTLY match ---- But currently they don't always, for example in the case below:
```
response_template = "### response:"
target = "This is a test"
formatted_prompt = f"Input ### response: {target}"
...
sliced_response_text = " This is a test"
```
Where we see their is an added space in the `slice_response_text`. While subtle and honestly hard to see often in the UI, for certain cases this off by one case can really screw things up.

*** Solution ***
The proposed solution is to get rid of having the user log `labels` / `targets` and instead directly infer them from as the `sliced_response_text`. This way, there will be no discrepancy between what the model sees and what is in the UI!

This change works quite well with the current system design because for computing token alignmnet we already `decode` the response tokens in the function `align_response_tokens_to_character_spans`. 

Other benefits are:
- We get out of the box the ability to show the `EOS` token and other special tokens, which we have wanted to make the default anyways! NOTE this is just for Decoder-Only modles.
- The user no longer needs to log labels! Just `Inputs` and `Formatted Prompt` 

***Demonstration***

Example run with the error: [run](https://console.dev.rungalileo.io/insights/aa9f6d64-4a53-4801-94b6-5b48e3f452cf/d22a7332-c831-4c05-9c32-744f1cc7b4b9?groupedBy=galileo_formatted_prompts&taskType=8)

If you look at the token DEPs you can see that the highlighted tokens don't make sense / are clearly off by one. The tokenizer would never encoder `{"m` as one token. It really means to encode ` {"` but the space is missing!

Example run with the fix: [run](https://console.dev.rungalileo.io/insights/8b403041-42ea-4227-9bbd-47c87e7f818a/99242382-8407-4f21-bc58-8658efd4bdf9?dataframeColumns=bleu%3B%26data_error_potential%3B%26generated_output%3B%26generated_token_logprobs%3B%26generated_uncertainty%3B%26input%3B%26perplexity%3B%26rouge%3B%26target&groupedBy=galileo_formatted_prompts&taskType=8)

In the first sample you can see that ` {"` is properly tokenized! Also looking at the other samples you see that the space appears before tokens NOT after. This was another sign that in retrospect I should have seen with other runs but did not notice!

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -31,7 +31,7 @@ If you want to train without a model, you can use the auto framework:
 """
 
 
-__version__ = "1.4.0"
+__version__ = "1.4.1"
 
 import sys
 from typing import Any, List, Optional
--- a/dataquality/loggers/data_logger/seq2seq/formatters.py
+++ b/dataquality/loggers/data_logger/seq2seq/formatters.py
@@ -43,7 +43,7 @@ class BaseSeq2SeqDataFormatter(ABC):
         tokenizer: PreTrainedTokenizerFast,
         max_tokens: Optional[int],
         split_key: str,
-    ) -> Tuple[AlignedTokenData, List[List[str]], Optional[List[str]]]:
+    ) -> Tuple[AlignedTokenData, List[List[str]], List[str]]:
         """Tokenize and align the `text` samples
 
         `format_text` tokenizes and computes token alignments for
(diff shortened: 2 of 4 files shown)


jonathangomesselman:
https://app.shortcut.com/galileo/story/10287

elboy3 (approved):

PR #822, opened 2023-12-26
feat: generation peft

jonathangomesselman:

Addresses two main issues:
1. Allowing for generation with `PEFTModels` from hf. This will be nearly the majority use case when people are fine-tuning very large models like Llama-2, since the general paradigm is to apply Parameter Efficient FT methods such as LORA and QLORA. Annoyingly a `PEFTModel`, while related to the hf library, comes from its own package `peft` and is not a subclass of a `PreTrainedModel`.
2. When generating, setting `model.config.use_cache=True` is very important for speed purposes. Otherwise generation is prohibitively slow!

### DIFF
--- a/dataquality/integrations/seq2seq/core.py
+++ b/dataquality/integrations/seq2seq/core.py
@@ -172,8 +172,10 @@ def watch(
 
     # A model of the correct type is required if we need to generate
     if generation_splits:
-        assert isinstance(model, (PreTrainedModel, PeftModel))
+        assert isinstance(
+            model, (PreTrainedModel, PeftModel)
         ), "model must be an instance of transformers PreTrainedModel"
+
         assert (
             model.can_generate()
         ), "model must contain a `generate` method for seq2seq"
--- a/dataquality/integrations/seq2seq/core.py
+++ b/dataquality/integrations/seq2seq/core.py
@@ -172,8 +172,7 @@ def watch(
 
     # A model of the correct type is required if we need to generate
     if generation_splits:
-        assert isinstance(model, PreTrainedModel) or isinstance(
-            model, PeftModel
+        assert isinstance(model, (PreTrainedModel, PeftModel))
         ), "model must be an instance of transformers PreTrainedModel"
         assert (
             model.can_generate()
--- a/dataquality/integrations/seq2seq/core.py
+++ b/dataquality/integrations/seq2seq/core.py
@@ -1,6 +1,7 @@
 from typing import List, Optional, Union
 from warnings import warn
 
+from peft import PeftModel
 from tokenizers import Tokenizer
 from transformers import GenerationConfig, PreTrainedModel, PreTrainedTokenizerFast
 
@@ -171,8 +172,8 @@ def watch(
 
     # A model of the correct type is required if we need to generate
     if generation_splits:
-        assert isinstance(
-            model, PreTrainedModel
+        assert isinstance(model, PreTrainedModel) or isinstance(
+            model, PeftModel
         ), "model must be an instance of transformers PreTrainedModel"
         assert (
             model.can_generate()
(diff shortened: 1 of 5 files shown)


dataquality/integrations/seq2seq/core.py
```diff
@@ -171,8 +172,8 @@ def watch(
 
     # A model of the correct type is required if we need to generate
     if generation_splits:
-        assert isinstance(
-            model, PreTrainedModel
+        assert isinstance(model, PreTrainedModel) or isinstance(
+            model, PeftModel
```
elboy3, line 176:
```suggestion
        assert isinstance(model, (PreTrainedModel, PeftModel))
```

https://stackoverflow.com/questions/33311258/python-check-if-variable-isinstance-of-any-type-in-list

dataquality/utils/seq2seq/generation.py
```diff
@@ -142,6 +142,11 @@ def add_generated_output_to_df(
         Updated Dataframe with the generated columns added (see above)
     """
     model.eval()
+    # During generation it is important to cache computation
```
elboy3, line 145: can you include the `why` here to give context for other developers! what is cached and why is it so much faster

dataquality/loggers/logger_config/seq2seq/seq2seq_base.py
```diff
@@ -15,7 +16,7 @@ class Seq2SeqLoggerConfig(BaseLoggerConfig):
     max_target_tokens: Optional[int] = None
     # For each split/inference-name, store sample id -> List[token_id] for the label
     id_to_tokens: Dict[str, Dict[int, List[int]]] = defaultdict(dict)
-    model: Optional[PreTrainedModel] = None
+    model: Optional[Union[PreTrainedModel, PeftModel]] = None
```
elboy3, line 19: just a thought - we could add a new `GalileoGenerationModel` type or something that is the union of multiple model types that we can keep adding to. this will make typing and isinstance simpler

dataquality/integrations/seq2seq/core.py
jonathangomesselman: Nice!

dataquality/loggers/logger_config/seq2seq/seq2seq_base.py
jonathangomesselman: That is a good point!

jonathangomesselman: Do you want to include that now?

dataquality/utils/seq2seq/generation.py
```diff
@@ -142,6 +142,17 @@ def add_generated_output_to_df(
...
     """
     model.eval()
+    # When generating it is important to set `use_cache = True`.
+    # - WHAT? Caching stores intermediate token activations / representations.
+    #   During autoregressive generation, the cache is updated each time a token
+    #   is generated.
+    # - WHY? Caching prevents re-computing token information during auto-regressive
+    #   generation, DRAMATICALLY speeding up performance. Every time a new token is
+    #   generated, we only need to do the forward pass for a single new token, as we
+    #   leverage the cached information to compute transformer based attention.
```
elboy3, line 152: nice. what's the downside, why wouldn't a user always use caching for this?

dataquality/loggers/logger_config/seq2seq/seq2seq_base.py
elboy3: i think it's a nice to have not a must have, your call

elboy3 (approved):

elboy3:
You'll have to bump the `__init__.py` patch version too to let customers access this

jonathangomesselman:
> You'll have to bump the `__init__.py` patch version too to let customers access this

Do I do this manually?

dataquality/utils/seq2seq/generation.py
jonathangomesselman: You never want to train with this setting as it results in a decent amount of additional GPU storage! So almost always (with big models) you will turn this off during training. Here we just make sure it is turned back on for inference

jonathangomesselman: I guess I could add something about this?

PR #825, opened 2024-01-22
fix(seq2seq): Include action info in data logger errors

dcaustin33:
Changed to make the errors more explicit as I ran into some trouble while using the data logger.

### DIFF
--- a/dataquality/loggers/data_logger/seq2seq/seq2seq_base.py
+++ b/dataquality/loggers/data_logger/seq2seq/seq2seq_base.py
@@ -191,6 +191,14 @@ class Seq2SeqDataLogger(BaseGalileoDataLogger):
         meta: Union[List[str], List[int], None] = None,
     ) -> None:
         """Helper to log a pandas or vaex df"""
+        assert S2SIC.input in df, (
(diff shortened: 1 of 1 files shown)


setu4993 (approved):

dataquality/loggers/data_logger/seq2seq/seq2seq_base.py
```diff
@@ -191,6 +191,14 @@ def _log_df(
         meta: Union[List[str], List[int], None] = None,
     ) -> None:
         """Helper to log a pandas or vaex df"""
+        assert S2SIC.input in df, (
+            f"Input column {S2SIC.input} not found in dataframe. "
+            f"Please add a column titled {S2SIC.input} to your dataframe "
+        )
```
elboy3, line 197: couldn't they include a column not named input and have the column_mapping map to input? i think the assertion error here needs to be dynamic / not sure this validation is in the right place

elboy3:
@dcaustin33 What error were you getting when you didn't include ID? can you post the stack trace (it's in minio as an error.log file, we can pair to find it)

PR #826, opened 2024-01-23
feat: Upload model after finishing training

franz101:
This allows inference and auto training

Part of:
https://app.shortcut.com/galileo/story/10988/jpmc-allow-simple-upload-of-csv-to-train-a-model-and-get-dataquality-insights


dataquality/utils/auto_trainer.py
```diff
@@ -24,5 +24,5 @@ def do_train(
     for inf_name in inf_names:
         dq.set_split(Split.inference, inference_name=inf_name)
         trainer.predict(test_dataset=encoded_data[inf_name])
-    dq.finish(wait=wait, create_data_embs=create_data_embs)
+    dq.finish(wait=wait, create_data_embs=create_data_embs, upload_model=True)
```
elboy3, line 27: auto is going to always upload model? idk why that feels too aggressive, it should be just when they specify it

dataquality/integrations/transformers_trainer.py
```diff
@@ -337,6 +337,16 @@ def watch(
...
+    helper_data["model"] = trainer.model
+    helper_data["model_parameters"] = {
+        "classifier_layer": classifier_layer,
+        "embedding_dim": embedding_dim,
+        "logits_dim": logits_dim,
+        "embedding_fn": embedding_fn,
+        "logits_fn": logits_fn,
+        "last_hidden_state_layer": last_hidden_state_layer,
+    }
+    helper_data["model_kind"] = "transformers"
```
elboy3, line 349: can we make a dataclass that includes `transformers` rather than magic str

dataquality/clients/api.py
```diff
@@ -895,3 +895,28 @@ def upload_file_for_project(
...
+        """
+        Returns information about the model for a given run.
+        Will also update the status to complete.
+        :param project_id: The project id
+        :param run_id: The run id
+        """
+        return self.make_request(
+            RequestType.GET,
+            url=f"{config.api_url}/{Route.projects}/{str(project_id)}/{Route.runs}/{str(run_id)}/{Route.model}",
+        )
```
elboy3, line 922: to me it's a bit odd that one of these endpoints is `upload_model` and the other `model`, shouldn't they be the same but one is a POST and one is a GET?

dataquality/utils/upload_model.py
```diff
@@ -0,0 +1,67 @@
...
+def upload_to_minio_using_presigned_url(presigned_url: str, file_path: str) -> Tuple:
+    """
+    Uploads a file to a presigned url.
+    """
+    with open(file_path, "rb") as f:
+        response = requests.put(presigned_url, data=f)
+        return response.status_code, response.text
+
+
+def upload_model_to_dq() -> None:
```
elboy3, line 36: i think this fn should take in a model 

also you are not upoading the model to dq you are uploading it to object store

dataquality/clients/api.py
```diff
@@ -895,3 +895,28 @@ def upload_file_for_project(
             },
         )
         return res
+
+    def get_presigned_url_for_model(
```
elboy3, line 899: this fn should just return the `presigned_url` based on the fn name

elboy3 (changes requested):
looking pretty good, but left a few suggestions, also gotta format / lint. Main concern is that I'm not sure all auto flows should upload model by default, would love to discuss that detail

franz101:
I agree with auto. We can just make sure the dq.auto on sagemaker always uploads so you can later use it for inference

dataquality/utils/auto_trainer.py
franz101: Otherwise you could not press the inference button. Let's disable it for now and then in the future, we can enable it.

dataquality/integrations/transformers_trainer.py
franz101: done

dataquality/clients/api.py
franz101: changed

dataquality/utils/upload_model.py
franz101: done, model is Any for now since huggingface can have many different model classes

franz101:
@elboy3 Sir, please reconsider (I have added all your requests)

dataquality/clients/api.py
```diff
@@ -895,3 +895,28 @@ def upload_file_for_project(
...
+        return self.make_request(
+            RequestType.POST,
+            url=f"{config.api_url}/{Route.projects}/{str(project_id)}/{Route.runs}/{str(run_id)}/{Route.model}",
+            body={"kind": model_kind, "parameters": model_parameters},
+        )["upload_url"]
+
+    def get_uploaded_model_info(self, project_id: UUID4, run_id: UUID4) -> Any:
+        """
+        Returns information about the model for a given run.
+        Will also update the status to complete.
```
elboy3, line 915: why does it update the status to complete? 

also what does it return, the model or a presigned url to download it?

dataquality/core/finish.py
```diff
@@ -85,6 +90,19 @@ def finish(
         f"Job {res['job_name']} successfully submitted. Results will be available "
         f"soon at {res['link']}"
     )
+    if upload_model:
+        try:
+            helper_data = get_model_logger().logger_config.helper_data
+            model = helper_data["model"]
+            model_parameters = helper_data["model_parameters"]
+            model_kind = helper_data["model_kind"]
```
elboy3, line 98: you should move these under the if on line 99. if "model" is not in `helper_data` it will have already thrown an error on 96

dataquality/clients/api.py
franz101: since we don't know on the backend when a minio upload is completed we update it every time we get the model and if the filename is not saved.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 license = {text = 'See LICENSE'}
 requires-python = ">=3.8"
 dependencies = [
-    "pydantic>=2.0.0",
```
elboy3, line 16: i don't think you need any of these.... i fixed the tests and all the formatting issues earlier today just merge the latest from main and keep the pyproject.toml the same

tests/loggers/test_seq2seq.py
```diff
@@ -647,6 +647,7 @@ def test_create_data_embs_df_custom_column(
 
     # Check that no exception is thrown and that data embs are created
     assert "text" not in df.get_column_names()
+
```
elboy3, line 650: don't need this whitespace

dataquality/core/finish.py
elboy3: wait, this isn't resolved. Can you move these below the if?

dataquality/clients/api.py
elboy3: so what status is it updating, the job status?

franz101: just the upload is completed. at first it's the put link, once it's put we will pull the download link from minio and add it to the entry

franz101:
Added test and reverted pyproject, ready to merge

elboy3 (approved):

PR #830, opened 2024-02-02
fix: re-add og internal reprocess fn

elboy3:
Re-introducing the old reprocess run logic since it was useful for transferred runs that don't have a preexisting job

### DIFF
--- a/dataquality/__init__.py
+++ b/dataquality/__init__.py
@@ -30,7 +30,6 @@ If you want to train without a model, you can use the auto framework:
         dataquality.get_insights()
 """
 
-
 __version__ = "1.6.0"
 
 import sys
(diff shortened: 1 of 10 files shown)


setu4993 (approved):
Nice.

setu4993 (dismissed):
Hmm... What changed about our `pre-commit` config that caused these changes here?

elboy3:
@setu4993 DQ doesn't have pre-commit set up, I can add that. But also I'm not sure, suddenly had a few failing tests and format issues, I think a new version of black just got pushed a few days ago. 

I can add pre-commit but also having poetry would be nice to avoid these mismatches between local and GH envs

setu4993 (approved):

PR #833, opened 2024-02-16
chore(deps): Migrate dependency management to use `poetry`

setu4993:
Switching over dependency management to `poetry`!


pyproject.toml
```diff
@@ -1,130 +1,121 @@
...
+setfit = {version ="==0.7.0", optional = true}
+
+# Cuda.
+ucx-py-cu11 = { version ="^0.36", source = "nvidia", optional = true }
+rmm-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
+raft-dask-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
+pylibraft-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
+dask-cudf-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
+cudf-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
+cuml-cu11 = {version = "^24.2.0", source = "nvidia", optional=true}
```
elboy3, line 60: Would be nice to have these as an optional extra, is that possible with poetry?

elboy3 (commented):
Looking really good, heroic stuff setu! Your PR description says WIP, what's left?

setu4993:
@elboy3 :

> Looking really good, heroic stuff setu! Your PR description says WIP, what's left?

Yup, it was WIP when I first created it. What I'm still waiting for is confirmation that we can drop Python 3.8 support.

elboy3:
@setu4993 sounds good. We have confirmation from Legartis, Reddit, and Musixmatch thus far. 

I will confirm with the remaining customers after my 1-1 with yash today

elboy3:
@setu4993 sounds good, will look into that later today

franz101:
How did you solve the Cuda issues?

setu4993:
> How did you solve the Cuda issues?

Used the latest versions :).

LICENSE
```diff
@@ -1,7 +1,201 @@
-Copyright 2021 Galileo Technologies, Inc.
+                                 Apache License
```
setu4993, line 1: Our license was MIT, which doesn't seem correct for what we are supporting. It doesn't even offer trademark protections!

pyproject.toml
```diff
@@ -1,131 +1,122 @@
...
-]
-minio = ["minio>=7.1.0,<7.2.0"]
-setfit = ["setfit==0.7.0"]
+license = "Apache-2.0"
+homepage = "https://github.com/rungalileo/dataquality"
+documentation = "https://rungalileo.gitbook.io/galileo"
+scripts = { dqyolo = "dataquality.dqyolo:main" }
+
+[tool.poetry.dependencies]
+python = "^3.9,<3.12"
```
elboy3, line 17: shouldn't this be <3.11. i don't think vaex and 3.11 are compatible yet

.github/workflows/test.yaml
```diff
@@ -13,34 +13,32 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
+          cache-dependency-path: "pyproject.toml"
 
       - name: install invoke
-        run: pip install invoke
+        run: pipx install invoke
 
       - name: install dependencies
-        run: inv install --no-editable
+        run: invoke install
```
elboy3, line 36: why this?

elboy3 (approved):
Looks great except i think we should also just restrict 3.9 and 3.10 in poetry file

.github/workflows/test.yaml
setu4993: I removed the conditional from `tasks.py`, so this makes it more aligned.

PR #837, opened 2024-03-02
chore(pre-commit): Update ruff, isort, black config, apply to all files

setu4993:
It really irked me that the `pre-commit` action was failing, so I just set the config correctly and applied it.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -7,4 +7,10 @@ on:
 
 jobs:
   pre-commit:
-    uses: rungalileo/.github/.github/workflows/pre-commit.yaml@main
+    runs-on: ubuntu-latest
+    env:
+      SKIP: "mypy"
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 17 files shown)


.github/workflows/pre-commit.yaml
```diff
@@ -7,4 +7,10 @@ on:
...
 jobs:
   pre-commit:
-    uses: rungalileo/.github/.github/workflows/pre-commit.yaml@main
+    runs-on: ubuntu-latest
+    env:
+      SKIP: "mypy"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+      - uses: pre-commit/action@v3.0.1
```
setu4993, line 16: We can't use the internal repo's source action since the access there is limited to private repos:

> Workflows in other repositories that are part of the 'rungalileo' organization can access the actions and reusable workflows in this repository. Access is allowed only from private repositories.

elboy3 (approved):

quinn-galileo (approved):

PR #841, opened 2024-03-05
ci: Fix pre-commit and stale workflows to remove mixup

setu4993:
**Description:** In my attempt to remove the dependence on internal workflows in #838, I mixed up and made changes to the workflow from #837. This should bring them back into alignment as we expect them to be.

Oops!


elboy3 (approved):
whoops

PR #845, opened 2024-03-18
feat(cu12): Support for cuda12

franz101:
**Shortcut:**
https://app.shortcut.com/galileo/story/12155/robustify-csv-upload-auto-training
**Description:**
Currently pip install dataquality[cuda] will not work on colab nor our csv upload

Some background colab is running cuda 12 (see screenshot in the comment)

Installing a pycuda 11 will cause an error and not execute running the umap locally

Reproducible examples (make sure to disconnect and delete runtime before testing it):
https://colab.research.google.com/drive/1_ExNBwu80aOriIO8YO1UDYsOREhRzbdO?usp=sharing


pyproject.toml
```diff
@@ -103,7 +103,7 @@ fastai = ">=2.7.11"
 portalocker = "==2.7.0"
 types-PyYAML = "==6.0.12.9"
 setfit = "==0.7.0"
-accelerate = ">=0.19.0"
+accelerate = "==0.27.2"
```
setu4993, line 106: Why pin this?

dataquality/clients/objectstore.py
```diff
@@ -49,7 +49,7 @@ def create_minio_client_for_exoscale_cluster(self) -> Any:
             "API Key."
         )
         return Minio(
-            endpoint=config.minio_fqdn,
+            endpoint=str(config.minio_fqdn),
```
setu4993, line 52: Curious about the change here?

setu4993 (commented):
Can you add a bit more context about why this change?

dataquality/utils/patcher.py
```diff
@@ -42,7 +42,10 @@ def __del__(self) -> None:
         """Call the cleanup function when the object is garbage collected."""
         # This will be called when the object is garbage collected.
         # However, it's not guaranteed to be called in all situations.
-        self.cleanup_manager._cleanup_callback(weakref.ref(self))
+        try:
+            self.cleanup_manager._cleanup_callback(weakref.ref(self))
+        except TypeError:
+            pass
```
elboy3, line 48: What would cause the TypeError and why do we want to do nothing on it

franz101:
Some background colab is running coda 12

Screenshot 2024-03-18 at 19 41 03


Installing a pycuda 11 will cause an error and not execute running the umap locally

https://colab.research.google.com/drive/1_ExNBwu80aOriIO8YO1UDYsOREhRzbdO?usp=sharing

dataquality/utils/patcher.py
franz101: The reference is already garbage collected and then this would call the function even though it does not exist anymore, I can look into this but this would cause the program to crash at the end after training is complete (training / upload is not affected)

dataquality/clients/objectstore.py
franz101: mypy was giving me an error so changed this for precommit

pyproject.toml
franz101: https://github.com/huggingface/transformers/issues/29216

franz101:
> Can you add a bit more context about why this change?

So a customer upload 114k samples which took hours to process the umap on our server.
I noticed A: Cuda is broken on colab
I noticed B: Changing to a cuda runtime in the sagemaker container caused two minor bugs / accelerate + garbage collection.

I have tested this fixed version on colab and sagemaker

setu4993:
But what about customers that still use Cuda 11?

franz101:
> But what about customers that still use Cuda 11?

Then we need to split it in dataquality[cu11]

Can't support both afaik

dataquality/clients/objectstore.py
setu4993: What I was trying to get at is: What changed _now_ that is causing it to fail?

Anyway.

pyproject.toml
```diff
@@ -49,16 +49,24 @@ opencv-python = "<=4.8.1.78"
 pydantic-settings = ">=2.0.0"
 minio = {version = ">=7.1.0,<7.2.0", optional=true}
 setfit = {version ="==0.7.0", optional = true}
+sentencepiece = "^0.2.0"
```
setu4993, line 52: Should this be optional?

franz101: Not sure, was like this before

franz101: I assume by the position

franz101:
@setu4993 any idea on how to fix the poetry part?
I was not able to make it work having them not installed or resolved together:
specifically it causes:
```
Because no versions of cuml-cu11 match >24.2.0,<25.0.0
 and cuml-cu11 (24.2.0) depends on pylibraft-cu11 (==24.2.*), cuml-cu11 (>=24.2.0,<25.0.0) requires pylibraft-cu11 (==24.2.*).
And because no versions of pylibraft-cu11 match >24.2.0,<24.3.dev0
 and pylibraft-cu11 (24.2.0) depends on cuda-python (>=11.7.1,<12.0a0), cuml-cu11 (>=24.2.0,<25.0.0) requires cuda-python (>=11.7.1,<12.0a0) or pylibraft-cu11 (>=24.2.dev0,<24.2.0).
Because pylibraft-cu12 (24.2.0) depends on cuda-python (>=12.0,<13.0a0)
 and no versions of pylibraft-cu12 match >24.2.0,<24.3.dev0, pylibraft-cu12 (>=24.2.0,<24.3.dev0) requires cuda-python (>=12.0,<13.0a0).
Thus, if cuml-cu11 (>=24.2.0,<25.0.0) and pylibraft-cu12 (>=24.2.0,<24.3.dev0) then pylibraft-cu11 (>=24.2.dev0,<24.2.0).
And because cuml-cu12 (24.2.0) depends on pylibraft-cu12 (==24.2.*)
 and no versions of cuml-cu12 match >24.2.0,<25.0.0, if cuml-cu11 (>=24.2.0,<25.0.0) and cuml-cu12 (>=24.2.0,<25.0.0) then pylibraft-cu11 (>=24.2.dev0,<24.2.0) or pylibraft-cu12 (>=24.2.dev0,<24.2.0).
And because dataquality depends on both pylibraft-cu12 (^24.2.0) and cuml-cu12 (^24.2.0), cuml-cu11 (>=24.2.0,<25.0.0) requires pylibraft-cu11 (>=24.2.dev0,<24.2.0).
So, because dataquality depends on both pylibraft-cu11 (^24.2.0) and cuml-cu11 (^24.2.0), version solving failed.
```


```
[project.optional-dependencies]
cuda = ["ucx-py-cu11", "rmm-cu11", "raft-dask-cu11", "pylibraft-cu11", "dask-cudf-cu11", "cudf-cu11", "cuml-cu11"]
cuda12 = ["ucx-py-cu12", "rmm-cu12", "raft-dask-cu12", "pylibraft-cu12", "dask-cudf-cu12", "cudf-cu12", "cuml-cu12"]
```

pyproject.toml
setu4993: It looks like it's working with any of the latest versions? https://github.com/huggingface/transformers/issues/29216#issuecomment-2009050327

setu4993:
We could use [dependency groups](https://python-poetry.org/docs/managing-dependencies/#optional-groups).

franz101:
I have added a not so pretty replacement to unblock this ticket.

elboy3 (changes requested):
My suggestion is to support cuda12 but have instructions for how a customer can run this if they are using cuda 11 in our docs

franz101:
reverted to pip install dataquality[cuda12]

PR #847, opened 2024-04-11
feat(cuda): Add support for cuda 12 so as to not conflict with users workflows

dcaustin33:
**Shortcut:** Na

**Description:** Adding support for Cuda 12 and allowing users to download the version they are using in their current workflows. FAQ documentation will be updated when this is merged

**Tests:**

- [ ] Local.
- [ ] CI.


tyler-galileo (approved):

dcaustin33:
Currently works for both cuda11 and cuda12 - most people in the future will be using cuda 12 so think we should make that the default. Only problem being we cannot pin the cuda11 versions. @franz101 I realized late you were doing the same work but what do you think of this solution. Will wait for @elboy3 to be back before merging anything

franz101:
@dcaustin33 where you able to make it work?

elboy3 (changes requested):
This is very nice. I didn't realize you could add cuda11 to the extras without having them in the dependencies. 

We just gotta relock poetry lock file (sorry, I pushed a small change to that file)

And we gotta look into that one failing test. It looks like `self.model` may be reserved within that class. 

And then we can merge

franz101:
Very simple pin keras to 2.15 (right now it's at 3)

franz101 (approved):
Make sure keras is below 2.15 (right now it's on 3)
https://github.com/rungalileo/dataquality/blob/5e1ef42cecf43d159cb79211d31d7ab07540dfb9/pyproject.toml#L120

Here TensorFlow needs to be <= 2.15.0

elboy3 (approved):
looking good, very nice

PR #873, opened 2024-10-02
feat(colab) bump versions to avoid conflicts in google colab environment

franz101:
**Description:**
- Updating all dependencies so there are no conflicts with the environment of Google Colab.
- Updating vaex to support newer versions of Python

To do:
- Set upper version bound


ShuaiShao93 (approved):

setu4993:
@franz101 : Let's push this out until we have everything working? I'm also not sure if we _need_ to update this right now... Your choice.

franz101:
Tests are passing :)
https://github.com/vaexio/vaex/issues/2439

franz101:
Screenshot 2024-10-10 at 01 10 18

🥇  34 seconds no warnings and errors vs 1:34

pyproject.toml
```diff
@@ -20,23 +20,23 @@ requests = ">=2.25.1"
 types-requests = ">=2.25.2"
 pandas = ">=0.20.0"
 pyarrow = ">=5.0.0"
-vaex-core = "4.17.1"
-vaex-hdf5 = ">=0.12,<0.13"
+vaex-core = "4.18.1"
```
setu4993, line 23: Do we need an exact pin here?

pyproject.toml
```diff
@@ -20,23 +20,23 @@ requests = ">=2.25.1"
...
+vaex-hdf5 = ">=0.12"
 diskcache = ">=5.2.1"
 resource = ">=0.2.1"
 tqdm = ">=4.62.3"
 blake3 = ">=0.2.1"
 wrapt = ">=1.13.3"
 scipy = ">=1.7.0"
 cachetools = ">=4.2.4"
-importlib-metadata = "<6.0.1"
+importlib-metadata = ">=4.0.0,<9.0.0"
```
setu4993, line 32: Curious why we're changing this?

pyproject.toml
```diff
@@ -45,9 +45,9 @@ imagededup = ">=0.3.1,<0.3.2"
 pyjwt = ">=2.8.0"
 peft = "*"
 # Pin opencv for linting incompatibility
-opencv-python = "<=4.8.1.78"
+opencv-python = ">=4.5.0,<4.10.0.85"
```
setu4993, line 48: Do we need this narrow a pin?

pyproject.toml
```diff
@@ -118,7 +119,7 @@ freezegun = ">=1.2.2"
 coverage = { version = ">=7.0.5", extras = ["toml"] }
 pytest-cov = ">=4.0.0"
 scikit-learn = ">=1.0"
-tensorflow = "<=2.15.0"
+tensorflow = ">=2.15.0"
```
setu4993, line 122: Why do we have different constraints in runtime and test?

setu4993 (commented):
Asked a few questions inline. Let's not merge this until they're resolved, please. I can work with you next week to get it ready.

pyproject.toml
franz101: no, can be upgradable, i guess if numpy version is higher lower vaex versions would not work

pyproject.toml
franz101: taken from colab pip freeze

pyproject.toml
franz101: same here, sticking to what's in colab. we could probably do >=4.5.0,<5"
though im not very familiar with this package

pyproject.toml
setu4993: I don't think we should use what Colab does. That's an artificial constraint.

franz101: I would say colab is very LTS and has healthy state of libraries, while the upper bound should not be set by colab, def the lower bound if not setting it leads to package conflicts or errors like it currently does.

pyproject.toml
franz101: yup

PR #881, opened 2024-12-04
feat: Support filtering for reviewed only edits

vpogrebi:
Added `reviewed_only` optional argument to `metrics.get_edited_dataframe()` procedure and added logic to filter DataFrame when this argument is True.

https://app.shortcut.com/galileo/story/21961/api-nlp-export-parity-between-dq-get-dataframe-and-nlp-studio


dataquality/metrics.py
```diff
@@ -405,6 +408,11 @@ def get_edited_dataframe(
         tagging_schema=tagging_schema,
     )
     data_df = vaex.open(file_name)
+
+    if reviewed_only:
+        # [SHORTCUT-21961] Filter DataFrame to contain only reviewed edits
```
setu4993, line 413: Shortcut tickets are internal only so we don't need to reference them in code. The `git blame` should provide enough context to link back.

```suggestion
```

setu4993 (commented):
Looks like a good start. Can we add a test for this please?

PR #882, opened 2024-12-17
chore(deps): Create separate group for CV dependencies

setu4993:
**Shortcut:** https://app.shortcut.com/galileo/story/23104/

**Description:** `imagededup` is producing a lot of pain for new installs, so making it optional for CV installs only, alongside a couple other CV exclusive dependencies.

**Tests:**

- [x] Local import (`import dataquality as dq` with all extras uninstalled).
- [x] CI.


setu4993:
https://app.shortcut.com/galileo/story/23104/dq-remove-imagededup-dependency-for-dq

dmallon (approved):
nice!

bogdan-galileo (approved):
Thanks. That imagededup was terrible, it was downloading a small pytorch model at install time, and it created issues with customers which were not allowed to download models (P&G).

